PR #4, opened 2020-07-17
added tests for PECS task

KseniiaMakarova:
FYI: Travis won't pass because test class is designed to compile only in case of a proper student solution. Yet I haven't found any other way to handle this.

1. 
Screenshot 2020-07-17 at 17 05 59

Reasoning: We ask students to parameterise MachineService and to add implementation by themselves. I don't think that we should give a hint providing the proper MachineServiceImpl class signature. 

2.
Screenshot 2020-07-17 at 17 10 53

Reasoning: In order to test startWorking() method, the tests are designed in a way that they will compile if a student provides a right solution.


src/test/java/core/mate/academy/model/MachineServiceTest.java
```diff
@@ -0,0 +1,94 @@
...
+    public void getAll_trackType() {
+        List<Machine> actual = machineService.getAll(Track.class);
+        assertEquals("The list should contain elements of type Track",
+                Track.class, actual.get(0).getClass());
+    }
+
+    @Test
+    public void fill_bulldozerValue() {
+        List<Object> machines = new ArrayList<>();
+        machineService.fill(machines, new Bulldozer());
```
boroda4436, line 44: Можливо тут має бути не пустий масив. А наприклад хоча б два-три об'єкти
```
machines.add(null);
machines.add(null);
machines.add(null);
```

boroda4436: https://youtu.be/humASfXSA7c

KseniiaMakarova: fixed, sorry for misunderstanding the method concept

PR #11, opened 2020-08-10
solved

ilya449:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,7 +17,7 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+        if (Bulldozer.class.equals(type)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        }
+        if (Excavator.class.equals(type)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        if (Track.class.equals(type)) {
+            return new ArrayList<>(new TrackProducer().get());
+        }
+        return null;
```
Sofasmile, line 25: return `null` isn't a good practice

src/main/java/core/mate/academy/service/MachineService.java
ilya449: Ok, changed

src/main/java/core/mate/academy/service/MachineServiceImpl.java
ilya449: fixed

PR #12, opened 2020-08-10
hw11 is done

vladkuzya:


src/main/java/core/mate/academy/service/TrackMachineProducerImpl.java
```diff
@@ -0,0 +1,19 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Track;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TrackMachineProducerImpl implements MachineProducer {
```
RomanDubovskyi, line 7: let's parameterize MachineProducer with Track here, that will put obligation to parameterize get method with right type
and same in each impl

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,48 @@
...
+        if (type == Excavator.class) {
+            List<Excavator> excavatorList = new ExcavatorMachineProducerImpl().get();
+            return new ArrayList<>(excavatorList);
+        }
+        throw new IllegalArgumentException("Wrong class");
+    }
+
+    @Override
+    public void fill(List<? super Machine> list, Machine value) {
+        Collections.fill(list, value);
```
RomanDubovskyi, line 36: let's remove Collections.fill() method and use our own loop only

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
```suggestion
    List<T> getAll(Class<? extends T> type);
```
let's use T here as well

PR #18, opened 2020-08-10
added solution

matviichuk92:


src/main/java/core/mate/academy/service/BulldozerProduce.java
```diff
@@ -0,0 +1,15 @@
...
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProduce implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        bulldozers.add(new Bulldozer());
+        bulldozers.add(new Bulldozer());
```
AnaLanda, line 12: You could initialize the fields of your Bulldozer here since you have a constructor for this in Bulldozer.class, e.g. 
bulldozers.add(new Bulldozer("Cat", "Spain"));

src/main/java/core/mate/academy/service/ExcavatorProduce.java
```diff
@@ -0,0 +1,15 @@
...
+import core.mate.academy.model.Excavator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProduce implements MachineProducer<Excavator> {
+    @Override
+    public List<Excavator> get() {
+        List<Excavator> excavators = new ArrayList<>();
+        excavators.add(new Excavator());
+        excavators.add(new Excavator());
```
AnaLanda, line 12: Here, you could also initialize new Excavator() with field values, like in line 11-12 above

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -6,7 +6,7 @@
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
```
AnaLanda, line 9: Like Sofia said in the Q&A, we can parametrize MachineService with <T> and then use <? extends T> or <? super T>, so that we are not restricted by the class Machine

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
AnaLanda, line 20: If we parametrize MachineService with "T", we can use <? extends T> here

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
...
      * Fill the machines list with passed value
      * Replace the Object with parametrized value.
      * This method should be able to work well with any type of machine passed as 'value'
+     *
      * @param machines - list of machines to be filled with value
-     * @param value - any object of machine sub class
+     * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, Machine value);
```
AnaLanda, line 30: (List<? super T> machines, T value);

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
...
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of tracks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
AnaLanda, line 38: If we parameterize MachineService with "T", we can use <? extends T> here

src/main/java/core/mate/academy/service/TrackProducer.java
```diff
@@ -0,0 +1,15 @@
...
+import core.mate.academy.model.Track;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TrackProducer implements MachineProducer<Track> {
+    @Override
+    public List<Track> get() {
+        List<Track> tracks = new ArrayList<>();
+        tracks.add(new Track());
+        tracks.add(new Track());
```
AnaLanda, line 12: You could initialize new Track() with some unique field values here, too

src/main/java/core/mate/academy/service/MachineService.java
matviichuk92: thank for comments I will fix that

matviichuk92:
done

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,6 +5,14 @@
...
 public class Bulldozer extends Machine {
+    private String brand;
+    private String country;
+
+    public Bulldozer(String brand, String country) {
+        this.brand = brand;
+        this.country = country;
+    }
+
     public Bulldozer() {
```
mikhail771, line 16: По-моему, конструктор без параметров должен быть первым методом

RomanDubovskyi: constructors in general should be before methods it's not crucial, which one goes first

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        if (Bulldozer.class.equals(type)) {
+            return new ArrayList<>(new BulldozerProduce().get());
+        }
+        if (Excavator.class.equals(type)) {
+            return new ArrayList<>(new ExcavatorProduce().get());
+        }
+        if (Track.class.equals(type)) {
+            return new ArrayList<>(new TrackProducer().get());
+        }
+        return null;
```
RomanDubovskyi, line 25: it's  not quite good to return null since it can produce NPE, let's throw exception or return empty list here

matviichuk92:
done

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -22,7 +22,7 @@
         if (Track.class.equals(type)) {
             return new ArrayList<>(new TrackProducer().get());
         }
-        return null;
+        return new ArrayList<>();
```
RomanDubovskyi, line 25: FYI: you can use existing constant Collections.EMPTY_LIST

matviichuk92: got it

PR #23, opened 2020-08-10
added solution

MaliukDaria:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,44 @@
...
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Track;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
```
RomanDubovskyi, line 12:
```suggestion
public class MachineServiceImpl implements MachineService<Machine> {
```
in this case we want to parameterize class with specific Type

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,17 @@
...
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        for (int i = 0, n = 3; i < n; i++) {
```
Sofasmile, line 11: better move 3 to constant

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,44 @@
...
+            machines = new BulldozerProducer().get();
+            return (List<Machine>) machines;
+        }
+        if (type.equals(Excavator.class)) {
+            machines = new ExcavatorProducer().get();
+            return (List<Machine>) machines;
+        }
+        if (type.equals(Track.class)) {
+            machines = new TrackProducer().get();
+            return (List<Machine>) machines;
```
Sofasmile, line 26: do we need to use return at all `if` statement? 
can we use it only at the end of the method?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
MaliukDaria: done

src/main/java/core/mate/academy/service/MachineServiceImpl.java
MaliukDaria: done

src/main/java/core/mate/academy/service/BulldozerProducer.java
MaliukDaria: done

PR #25, opened 2020-08-10
homeWork11

khOksana99:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+        }
+        if (type.equals(Excavator.class)) {
+            List<Excavator> excavators = new ExcavatorProducer().get();
+            return new ArrayList<>(excavators);
+        }
+        if (type.equals(Track.class)) {
+            List<Track> tracks = new TrackProducer().get();
+            return new ArrayList<>(tracks);
+        }
+        return null;
```
Sofasmile, line 29: return `null` isn't a good practice

src/main/java/core/mate/academy/service/MachineService.java
khOksana99: done

src/main/java/core/mate/academy/service/MachineServiceImpl.java
khOksana99: done

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
...
      * Fill the machines list with passed value
      * Replace the Object with parametrized value.
      * This method should be able to work well with any type of machine passed as 'value'
+     *
      * @param machines - list of machines to be filled with value
-     * @param value - any object of machine sub class
+     * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, T value);
```
Sofasmile, line 30: how about other methods? let's fix all methods in this interface

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+        }
+        if (type.equals(Excavator.class)) {
+            List<Excavator> excavators = new ExcavatorProducer().get();
+            return new ArrayList<>(excavators);
+        }
+        if (type.equals(Track.class)) {
+            List<Track> tracks = new TrackProducer().get();
+            return new ArrayList<>(tracks);
+        }
+        throw new NullPointerException("type can't be null");
```
Sofasmile, line 29: better return some other type of exception, it can be IllegalArgumentException

khOksana99: done

src/main/java/core/mate/academy/service/MachineService.java
khOksana99: done

PR #26, opened 2020-08-10
hw10

Alkey:


src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -1,6 +1,8 @@
 package core.mate.academy.model;
 
-public abstract class Machine implements Workable {
+import core.mate.academy.service.MachineProducer;
+
+public abstract class Machine implements Workable, MachineProducer<Machine> {
```
Sofasmile, line 5: do we need to implement MachineProducer here?
let's create separate classes for this purpose

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,13 @@
...
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
+     *
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends Machine> get();
```
Sofasmile, line 12: let's use T instead of Machine

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```

PR #29, opened 2020-08-10
Hm9

floriangeyer21:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,49 @@
...
+        }
+        if (type.equals(Bulldozer.class)) {
+            List<Bulldozer> bulldozerList = new BulldozerProducer().get();
+            return new ArrayList<>(bulldozerList);
+        }
+        if (type.equals(Track.class)) {
+            List<Track> trackList = new TrackProducer().get();
+            return new ArrayList<>(trackList);
+        }
+        return null;
```
RomanDubovskyi, line 28: it's not quite good to return null since it can produce NPE, let's throw exception or return empty list here

PR #31, opened 2020-08-10
Hw11

vremere-a:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
```
Sofasmile, line 5: let's add parent class Machine here:
<T extend Machine>

vremere-a:
done

src/main/java/core/mate/academy/service/MachineProducer.java
vremere-a: done

PR #32, opened 2020-08-10
ready solution hw-11

RomanHreits:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,50 @@
...
+        }
+        if (type == Bulldozer.class) {
+            MachineProducer producer = new BulldozerProducer();
+            return producer.get();
+        }
+        if (type == Excavator.class) {
+            MachineProducer producer = new ExcavatorProducer();
+            return producer.get();
+        }
+        return null;
```
RomanDubovskyi, line 28: it's not quite good to return null since it can produce NPE, let's throw exception or return empty list here

RomanHreits: done...added throw Exception

PR #33, opened 2020-08-10
hw.11

OlessiaKharchenko:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(Excavator.class)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else if (type.equals(Bulldozer.class)) {
```
Sofasmile, line 18: do we need to use else-if after return statement?

PR #36, opened 2020-08-11
Homework 11 completed

Borisko130:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,6 +5,10 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+
+    private int weight;
+    private int maxScoopWeight;
```
RomanDubovskyi, line 10: how can i get access to these fields?

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,7 +17,7 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
```suggestion
    List<T> getAll(Class<? extends T> type);
```
let's do it this way, so we would have consistent code in this class

src/main/java/core/mate/academy/model/Bulldozer.java
Borisko130: I've added getters and setters for these fields

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,17 @@
...
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> list = new ArrayList<>();
+        for (int i = 0; i < 3; i++) {
```
Sofasmile, line 12: better move 3 to constant

PR #37, opened 2020-08-11
solution added

Vosker:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        if (Bulldozer.class.equals(type)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        }
+        if (Excavator.class.equals(type)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        if (Track.class.equals(type)) {
+            return new ArrayList<>(new TrackProducer().get());
+        }
+        return null;
```
RomanDubovskyi, line 26: null may produce NPE, so it would be more safe to return Collections.EMPTY_LIST

PR #39, opened 2020-08-11
add solution

Victoria-Moiseienko:


src/main/java/core/mate/academy/service/BuldozerProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BuldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List get() {
```
Sofasmile, line 9: how about parametrization?

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    @Override
+    public List get() {
```
Sofasmile, line 9: the same

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,22 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
-
+    List<Machine> getAll(Class<? extends Machine> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Track;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
```
Sofasmile, line 12: do we need to parametrize class with Implementation?
let's do it only with interface

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        if (type == Bulldozer.class) {
+            return new BuldozerProducer().get();
+        }
+        if (type == Excavator.class) {
+            return new ExcavatorProducer().get();
+        }
+        if (type == Track.class) {
+            return new TrackProducer().get();
+        }
+        return null;
```
Sofasmile, line 24: return `null` isn't a good practice

Victoria-Moiseienko: Done

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Victoria-Moiseienko: Done.

src/main/java/core/mate/academy/service/MachineService.java
Victoria-Moiseienko: Done

src/main/java/core/mate/academy/service/BuldozerProducer.java
Victoria-Moiseienko: Done

src/main/java/core/mate/academy/service/ExcavatorProducer.java
Victoria-Moiseienko: Done

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BuldozerProducer().get());
+        }
+        if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        if (type == Track.class) {
+            return new ArrayList<>(new TrackProducer().get());
+        }
+        throw new ClassCastException("Unknown class [" + type + "]");
```
RomanDubovskyi, line 25: ClassCastException won't be quite good here, cause it appears while casting of invalid class, for example: 
Human fred = (Human) new Object();
```suggestion
        throw new RuntimeException("Unknown class [" + type + "]");
```

Victoria-Moiseienko: Done

PR #48, opened 2020-08-11
Home Work11

KostiantynVorobiov:


src/main/java/core/mate/academy/service/BulldozerMachineProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerMachineProducer implements MachineProducer<Bulldozer> {
+    public List<Bulldozer> get() {
```
RomanDubovskyi, line 8: FYI: 
it's better to use @Override annotation if you implementing interface method, that will make code more clear for reader

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,7 +17,7 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
RomanDubovskyi, line 20: use T with this method as well

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerMachineProducer().get());
+        }
+        if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorMachineProducer().get());
+        }
+        if (type == Track.class) {
+            return new ArrayList<>(new TrackMachineProducer().get());
+        }
+        return null;
```
RomanDubovskyi, line 25: it's not quite good to return null since it can produce NPE, let's throw exception or return empty list here

src/main/java/core/mate/academy/service/BulldozerMachineProducer.java
KostiantynVorobiov: Done

src/main/java/core/mate/academy/service/MachineService.java
KostiantynVorobiov: Done

src/main/java/core/mate/academy/service/MachineServiceImpl.java
KostiantynVorobiov: Done

PR #52, opened 2020-08-11
hw

YaTut1901:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +26,13 @@
      * @param machines - list of machines to be filled with value
      * @param value - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, T value);
```
Sofasmile, line 29: let's use T instead of Machine with all methods in this interface

PR #53, opened 2020-08-11
Hw 11

SetMyDream:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
...
      * Fill the machines list with passed value
      * Replace the Object with parametrized value.
      * This method should be able to work well with any type of machine passed as 'value'
+     *
      * @param machines - list of machines to be filled with value
-     * @param value - any object of machine sub class
+     * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, T value);//? super Machine
```
```suggestion
    void fill(List<? super T> machines, T value);
```
you wrote correct, let's use T here, so we will stick to the Type assigned during class declaration

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
RomanDubovskyi, line 20: same here, let's use T type

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +17,23 @@
...
+    void fill(List<? super T> machines, T value);//? super Machine
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of tracks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
RomanDubovskyi, line 38: let's use T

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        if (Bulldozer.class.equals(type)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        }
+        if (Excavator.class.equals(type)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        if (Track.class.equals(type)) {
+            return new ArrayList<>(new TrackProducer().get());
+        }
+        return null;
```
RomanDubovskyi, line 26: it's not very good to return null, because it may generate NPE, let's return empty list or throw exception here

src/main/java/core/mate/academy/service/MachineService.java
SetMyDream: i see

SetMyDream: Got no idea how to solve it through the <T>

SetMyDream: Done, but compiler complain that somtimes T are in conflict with Machine... so i let him just to re-write it everywhere it suggested.

src/main/java/core/mate/academy/service/MachineService.java
SetMyDream: done

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+import core.mate.academy.model.Track;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.NoSuchElementException;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<T> {
```
RomanDubovskyi, line 14:
```suggestion
public class MachineServiceImpl implements MachineService<Machine> {
```
we used T in MachineService class, while declaring abstract behaviour. 
Here we're implementing this logic, so it makes sense to use specific generic type Machine

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+        }
+        if (Track.class.equals(type)) {
+            return new ArrayList<>(new TrackProducer().get());
+        }
+        throw new NoSuchElementException("Sorry, looks like "
+                + type + " is not a Machine");
+    }
+
+    @Override
+    public void fill(List<? super T> machines, T value) {
```
RomanDubovskyi, line 32: same here in all methods

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -17,22 +16,23 @@
      * @param type - any class of Machine sub class. For example: Track.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class type);
+    List<core.mate.academy.model.T> getAll(Class<? extends T> type);
```
RomanDubovskyi, line 19:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,11 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T extends core.mate.academy.model.T> {
```
RomanDubovskyi, line 5:
```suggestion
public interface MachineProducer<T extends Machine> {
```

src/main/java/core/mate/academy/model/T.java
```diff
@@ -1,6 +1,6 @@
 package core.mate.academy.model;
 
-public abstract class Machine implements Workable {
+public abstract class T implements Workable {
```
RomanDubovskyi, line 3:
```suggestion
public abstract class Machine implements Workable {
```

PR #65, opened 2020-12-10
Pecs by oleh kravets

Ketch92:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
```
gigbat, line 10: Can we use only `<T>`?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.TruckProducer;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+    @Override
```
gigbat, line 15: It's up to you, you can do class without generics, but make into MachineService<T> generic to MachineService<Machine>

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer<? extends Machine> mp;
+        if (type != null && type.equals(Bulldozer.class)) {
```
gigbat, line 18: You can just compare two class using ==

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+            mp = new BulldozerProducer();
+        } else if (type != null && type.equals(Excavator.class)) {
```
gigbat, line 20: same as previous

gigbat (changes requested):

src/main/java/core/mate/academy/service/MachineService.java
Ketch92: yes, we can

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Ketch92: done

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Ketch92: done

gigbat (approved):

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
Sofasmile, line 9: if we use only `T` type, we can parameterize this interface with any type of class (It can be Person, String, Object...) Is it valid behavior for MachineService interface?
@gigbat be attentive))

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
Sofasmile, line 20: If class MachineServiceImpl will be parameterized by Truck. Can we pass Bulldozer.class to the method getAll()? Is it valid behavior?
the same with other methods

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+import core.mate.academy.model.ExcavatorProducer;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.TruckProducer;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
```
Sofasmile, line 14: we should move from abstraction to implementation. So we don't need to parameterize MachineServiceImpl

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer<? extends Machine> mp;
+        if (type == Bulldozer.class) {
+            mp = new BulldozerProducer();
+        } else if (type == Excavator.class) {
+            mp = new ExcavatorProducer();
+        } else {
+            mp = new TruckProducer();
+        }
+        return (List<Machine>) mp.get();
```
Sofasmile, line 25:
```suggestion
        if (type == Bulldozer.class) {
            return new ArrayList<>(new BulldozerProducer().get());
        } if (type == Excavator.class) {
            return new ArrayList<>(new ExcavatorProducer().get());
        } if (type == Truck.class) {
            return new ArrayList<>(new TruckProducer().get());
        }
        return Collections.emptyList();
```

Sofasmile (changes requested):

src/main/java/core/mate/academy/service/MachineService.java
gigbat: Okey

src/main/java/core/mate/academy/service/MachineService.java
Ketch92: I have been considering this. Thx

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Ketch92: On the lecture, Bohdan was implementing MachineServiceImpl and made it as I did. Then he explained that this is better than non-parametrized MachineServiceImpl cause we achieve better flexibility. I tend to leave my solution there

src/main/java/core/mate/academy/service/MachineService.java
Ketch92: You are right, thx)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Ketch92: Thanks, that is definitely better solution.

src/main/java/core/mate/academy/service/MachineService.java
Ketch92: Ok I got confused... see you on QnA

Ketch92: Resolved it with Roman

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -5,37 +5,40 @@
...
  */
-public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+public class MachineServiceImpl implements MachineService<Machine> {
     @Override
     public List<Machine> getAll(Class<? extends Machine> type) {
-        MachineProducer<? extends Machine> mp;
         if (type == Bulldozer.class) {
-            mp = new BulldozerProducer();
+            return new ArrayList<>(new BulldozerProducer().get());
         } else if (type == Excavator.class) {
```
Sofasmile, line 22: do we need else block after return statement?

Sofasmile (changes requested):
good job) Left one comment

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Ketch92: Fair enough!)

Sofasmile (approved):
awesome)

PR #67, opened 2020-12-10
Do all tasks

YuraMitryakhin:
Created class BulldozerProducer, ExcavatorProducer and TruckProducer which implements MachineProducer. And also realized class MachineServiceImpl.

### DIFF
--- a/src/main/java/core/mate/academy/service/BulldozerProducer.java
+++ b/src/main/java/core/mate/academy/service/BulldozerProducer.java
@@ -1,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        Bulldozer bulldozer = new Bulldozer("Cat", "Orange", 1000);
+        bulldozers.add(bulldozer);
+        return bulldozers;
+    }
+}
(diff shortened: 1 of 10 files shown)


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
```
IvanLuchkin, line 7: We need to parametrize `MachineProducer` interface to let him know that this implementation is dependent on the exact inheritant of `Machine` (same with other producers)

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
...
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        Bulldozer bulldozer = new Bulldozer("Cat", "Orange", 1000);
+        bulldozers.add(bulldozer);
```
IvanLuchkin, line 12: Redundant variable declaration can be removed (also applies to other producers)

IvanLuchkin (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        }
+        if (type == Excavator.class) {
+            machines = new ExcavatorProducer().get();
+            return new ArrayList<>(machines);
+        }
+        if (type == Truck.class) {
+            machines = new TruckProducer().get();
+            return new ArrayList<>(machines);
+        }
+        return null;
```
RomanDubovskyi, line 27:
```suggestion
        return Collections.EMPTY_LIST;
```

RomanDubovskyi (approved):

PR #71, opened 2020-12-12
create classes such as "BulldozerProducer", "ExcavatorProducer" and "…

YuraBondarenko:
…TruckProducer". Implemented "Bulldozer", "Truck", "Excavator" and "MachineServiceImpl". Also parameterize "MachineService" and other classes

### DIFF
--- a/src/main/java/core/mate/academy/model/Bulldozer.java
+++ b/src/main/java/core/mate/academy/model/Bulldozer.java
@@ -1,10 +1,14 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
+    private boolean hasWheels;
+
+    public Bulldozer(boolean hasWheels, String name, String color) {
+        this.hasWheels = hasWheels;
+        setName(name);
+        setColor(color);
+    }
+
     public Bulldozer() {
     }
 
--- a/src/main/java/core/mate/academy/service/BulldozerProducer.java
+++ b/src/main/java/core/mate/academy/service/BulldozerProducer.java
@@ -1,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozerList = new ArrayList<>();
+        bulldozerList.add(new Bulldozer(true, "Bulldozer1", "Yellow"));
+        bulldozerList.add(new Bulldozer(false, "Bulldozer2", "Blue"));
+        return bulldozerList;
+    }
+}
--- a/src/main/java/core/mate/academy/service/MachineService.java
+++ b/src/main/java/core/mate/academy/service/MachineService.java
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
 
     /**
      * Return the list of machines.
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class<? extends T> type);
 
     /**
      * Fill the machines list with passed value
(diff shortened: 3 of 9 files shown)


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
...
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozerList = new ArrayList<>();
+        bulldozerList.add(new Bulldozer(true, "Bulldozer1", "Yellow"));
+        bulldozerList.add(new Bulldozer(false, "Bulldozer2", "Blue"));
```
antoniukivan, line 12: add variable for every call of `new` operator

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class type);
```
antoniukivan, line 20: if you'll get class type and it won't a Mashine

antoniukivan (changes requested):

antoniukivan (approved):

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,10 +1,14 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
+    private boolean hasWheels;
```
Sofasmile, line 4: how will we get access to the field?

the same with other classes

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,10 +1,14 @@
...
+
+    public Bulldozer(boolean hasWheels, String name, String color) {
+        this.hasWheels = hasWheels;
+        setName(name);
+        setColor(color);
```
Sofasmile, line 9: we can create a constructor in the class Machine and use it here:
```
super(name, color);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, Machine value);
```
Sofasmile, line 30: Why do we use Machine here? not T type?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class type) {
+        if (type == Truck.class) {
+            TruckProducer truckProducer = new TruckProducer();
```
Sofasmile, line 14: we should operate with abstraction, not an implementation, so let's operate interfaces instead of class here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+            List<? extends Truck> trucks = truckProducer.get();
+            return new ArrayList<>(trucks);
```
Sofasmile, line 16:
```suggestion
            return new ArrayList<>(truckProducer.get());
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        } else if (type == Bulldozer.class) {
```
Sofasmile, line 17: do we need else block after return statement?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
+            List<? extends Bulldozer> bulldozers = bulldozerProducer.get();
+            return new ArrayList<>(bulldozers);
+        } else if (type == Excavator.class) {
+            ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+            List<? extends Excavator> excavators = excavatorProducer.get();
+            return new ArrayList<>(excavators);
+        } else {
+            return null;
```
Sofasmile, line 26: return null is a bad practice. Let's return an empty list instead

Sofasmile (changes requested):

Sofasmile (approved):
good job)

PR #72, opened 2020-12-12
я сделаль

Artificial-Friend:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
```
IvanLuchkin, line 9: We don't need to set any bounding here, because `MachineServiceImpl` parametrizes this interface with `Machine` type in its signature. If this bounding is set, our interface is depending on the exact implementation (which is working only with the `Machine` hierarchy), therefore, this interface is not reusable in any other part of the program.
```suggestion
public interface MachineService<T> {
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,47 @@
...
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+
+    @Override
+    public void fill(List<? super Machine> machines, Object value) {
```
IvanLuchkin, line 16: This implementation is depending on the `Machine` hierarchy already (see class signature). Is it safe to keep `Object` as possible parameter type?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,47 @@
...
+            machine.doWork();
+        }
+    }
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
+            List<Bulldozer> machines = bulldozerProducer.get();
+            return new ArrayList<>(machines);
```
IvanLuchkin, line 34: Redundant variables, can be simplified (as well as in other cases in this method)

IvanLuchkin (changes requested):

IvanLuchkin:
Also, please, consider improving your commit messages 😂

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,9 +5,28 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private int bulldozerSpeed;
+    private String bulldozerColor;
```
RomanDubovskyi, line 9:
```suggestion
    private int speed;
    private String color;
```

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,9 +5,28 @@
  * Do not remove no-args constructor
  */
 public class Excavator extends Machine {
+    private int excavatorSpeed;
+    private String excavatorColor;
```
RomanDubovskyi, line 9: same here and in other classes

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
```suggestion
    List<Machine> getAll(Class<? extends T> type);
```
this will make more flexible

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, T value);
```
```suggestion
    void fill(List<? super T> machines, T value);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@
...
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
RomanDubovskyi, line 38:
```suggestion
    void startWorking(List<? extends T> machines);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
```
RomanDubovskyi, line 13:
```suggestion
public class MachineServiceImpl implements MachineService<Machine> {
```
since it's impl we should declare specific parameter here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else if (type == Truck.class) {
+            return new ArrayList<>(new TruckProducer().get());
+        } else {
+            return null;
```
RomanDubovskyi, line 38:
```suggestion
        return Collections.EMPTY_LIST;
```

RomanDubovskyi (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if (type.equals(Excavator.class)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
```
RomanDubovskyi, line 21: we don't need else after return in if 
```suggestion
        } 
         if (type.equals(Excavator.class)) {
            return new ArrayList<>(new ExcavatorProducer().get());
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
RomanDubovskyi, line 9:
```suggestion
public interface MachineService<T extends Machine> {
```
your first impl was corrects because it's better restrict type to avoid MachineService<Cat>

RomanDubovskyi (approved):

PR #75, opened 2020-12-13
HM_PECS complete

grom4er:
Create all fields. Fill methods.

### DIFF
--- a/src/main/java/core/mate/academy/model/Truck.java
+++ b/src/main/java/core/mate/academy/model/Truck.java
@@ -8,6 +8,13 @@ public class Truck extends Machine {
     public Truck() {
     }
 
+    public TruckType getType() {
+        return type;
+    }
+
+    public void setType(TruckType type) {
+        this.type = type;
+    }
 
     public void doWork() {
 
--- a/src/main/java/core/mate/academy/model/Truck.java
+++ b/src/main/java/core/mate/academy/model/Truck.java
@@ -1,15 +1,16 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Truck
- * Do not remove no-field constructor
- */
+import core.mate.academy.model.enums.TruckType;
+
 public class Truck extends Machine {
+    private TruckType type;
+
     public Truck() {
     }
 
-    @Override
+
     public void doWork() {
+
         System.out.println("Truck started to work");
     }
 }
--- a/src/main/java/core/mate/academy/model/enums/Blade.java
+++ b/src/main/java/core/mate/academy/model/enums/Blade.java
@@ -1,0 +1,5 @@
+package core.mate.academy.model.enums;
+
+public enum Blade {
+    S_BLADE, U_BLADE, S_U_BLADE;
+}
--- a/src/main/java/core/mate/academy/service/MachineService.java
+++ b/src/main/java/core/mate/academy/service/MachineService.java
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
 
     /**
      * Return the list of machines.
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
 
     /**
      * Fill the machines list with passed value
@@ -26,7 +27,7 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, T value);
 
     /**
      * Call the method doWork() from each machine.
@@ -34,5 +35,5 @@ public interface MachineService {
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
 }
--- a/src/main/java/core/mate/academy/service/MachineServiceImpl.java
+++ b/src/main/java/core/mate/academy/service/MachineServiceImpl.java
@@ -1,7 +1,33 @@
 package core.mate.academy.service;
 
-/**
- * Your implementation of MachineService.
- */
-public class MachineServiceImpl {
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.getSimpleName().equals("Bulldozer")) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if (type.getSimpleName().equals("Excavator")) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else {
+            return new ArrayList<>(new TruckProducer().get());
+        }
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        for (int i = 0; i < 3; i++) {
+            machines.set(i, value);
+        }
+    }
+
+    @Override
+    public void startWorking(List<? extends Machine> machines) {
+        for (Machine machine : machines) {
+            machine.doWork();
+        }
+    }
 }
--- a/src/main/java/core/mate/academy/model/Bulldozer.java
+++ b/src/main/java/core/mate/academy/model/Bulldozer.java
@@ -1,11 +1,19 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
+import core.mate.academy.model.enums.Blade;
+
 public class Bulldozer extends Machine {
+    private Blade blade;
+
     public Bulldozer() {
+    }
+
+    public Blade getBlade() {
+        return blade;
+    }
+
+    public void setBlade(Blade blade) {
+        this.blade = blade;
     }
 
     @Override
--- a/src/main/java/core/mate/academy/model/Excavator.java
+++ b/src/main/java/core/mate/academy/model/Excavator.java
@@ -1,10 +1,17 @@
 package core.mate.academy.model;
 
-
 public class Excavator extends Machine {
     private int createYear;
 
     public Excavator() {
+    }
+
+    public int getCreateYear() {
+        return createYear;
+    }
+
+    public void setCreateYear(int createYear) {
+        this.createYear = createYear;
     }
 
     @Override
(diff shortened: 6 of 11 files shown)


boroda4436 (changes requested):
Travis failed

src/main/java/core/mate/academy/model/enums/Blade.java
```diff
@@ -0,0 +1,5 @@
+package core.mate.academy.model.enums;
+
+public enum Blade {
+    S_BLADE, U_BLADE, S_U_BLADE;
```
boroda4436, line 4: can we use more informative enum values? Now I don't understand what mean `S`, `U` or `S_U`

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -1,15 +1,23 @@
...
+
     public Truck() {
     }
 
-    @Override
+    public TruckType getType() {
+        return type;
+    }
+
+    public void setType(TruckType type) {
```
boroda4436, line 15: Are you using somewhere this method?

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
```
boroda4436, line 20:
```suggestion
    List<T> getAll(Class<? extends T> type);
```
It's better to return `T`. In this way, your code will be more flexible

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
boroda4436, line 9:
```suggestion
public interface MachineService<T extends Machine> {
```
You can bound your type

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@
...
+    void fill(List<? super T> machines, T value);
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
boroda4436, line 38:
```suggestion
    void startWorking(List<? extends T> machines);
```
If your `T` will be bounded at the class level you may use just `T`

boroda4436 (changes requested):

src/main/java/core/mate/academy/model/enums/Blade.java
grom4er: https://en.wikipedia.org/wiki/Bulldozer#Blade
Насколько я понял, это были их тех названия. Я сделал на описание, но считаю что старый вариант более лучше.

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
-/**
- * Parameterize this service and add its implementation in a separate class.
- */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
```
IvanLuchkin, line 6: If we generify our interface like this, it is dependent on the exact implementation (`MachineServiceImpl`), which works only with `Machine` hierarchy. Therefore, this particular interface is not reusable for any other type
```suggestion
public interface MachineService<T> {
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.getSimpleName().equals(TYPE_NAME_BULLDOZER)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if (type.getSimpleName().equals(TYPE_NAME_EXCAVATOR)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else if (type.getSimpleName().equals(TYPE_NAME_Truck)) {
+            return new ArrayList<>(new TruckProducer().get());
+        }
+        return new ArrayList<>();
```
IvanLuchkin, line 22: Isn't it better to return null?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
- */
-public class MachineServiceImpl {
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    private static final String TYPE_NAME_BULLDOZER = "Bulldozer";
+    private static final String TYPE_NAME_EXCAVATOR = "Excavator";
+    private static final String TYPE_NAME_Truck = "Truck";
```
IvanLuchkin, line 10: Why don't just use `getClass()` and get rid of these constants? What if we want to add another inheritant of `Machine`?

IvanLuchkin (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    private static final Class TYPE_BULLDOZER = Bulldozer.class;
+    private static final Class TYPE_EXCAVATOR = Excavator.class;
+    private static final Class TYPE_TRUCK = Truck.class;
```
IvanLuchkin, line 13: Mentor comment needed

IvanLuchkin (approved):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
grom4er: Мое объяснение:
Машин может быть N количество, в случае расширения/изменения метода fill, другому разработчику  проще понять почему именно эти 3 машины у меня на выбор.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
RomanDubovskyi: No, it's better to return empty list, because with null you may receive NPE

src/main/java/core/mate/academy/service/MachineService.java
RomanDubovskyi: image

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -1,15 +1,24 @@
...
+        super(name, color);
+        this.type = type;
+    }
+
+    public TruckType getType() {
+        return type;
+    }
+
     public void doWork() {
+
```
RomanDubovskyi, line 21: we don't need to add this extra line

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(TYPE_BULLDOZER)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if (type.equals(TYPE_EXCAVATOR)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else if (type.equals(TYPE_TRUCK)) {
+            return new ArrayList<>(new TruckProducer().get());
+        }
+        return null;
```
RomanDubovskyi, line 25:
```suggestion
        return Collections.EMPTY_LIST;
```
prev solution was ok, it's better to return empty list then null. This way user of your code won't need to write additional null check and will avoid NPE

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        for (int i = 0; i < COUNT_OF_FILL_LIST; i++) {
```
RomanDubovskyi, line 30:
```suggestion
        for (int i = 0; i < machines.size(); i++) {
```
if list has 100 elements we will always set first 3, and if list.size is 1 you will get IndexOutOfBound
https://mate-academy.github.io/jv-program-common-mistakes/java-core/pecs/pecs.html

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    private static final Class TYPE_BULLDOZER = Bulldozer.class;
+    private static final Class TYPE_EXCAVATOR = Excavator.class;
+    private static final Class TYPE_TRUCK = Truck.class;
+    private static final int COUNT_OF_FILL_LIST = 3;
```
RomanDubovskyi, line 14: we don't need this variable

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    private static final Class TYPE_BULLDOZER = Bulldozer.class;
+    private static final Class TYPE_EXCAVATOR = Excavator.class;
```
RomanDubovskyi, line 12: in future afte hashMap you will learn how to handle such case with a strategy pattern, that solution will be flexible. 
For now these constants don't give us a lot of practical value, we can remove them

RomanDubovskyi (changes requested):

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -2,37 +2,11 @@
 
 import java.util.List;
 
-/**
- * Parameterize this service and add its implementation in a separate class.
- */
-public interface MachineService {
+public interface MachineService<T> {
```
RomanDubovskyi, line 5:
```suggestion
public interface MachineService<T extends Machine> {
```
better to restrict usage of generics so you won't be able to user MachineService<Cat> etc

RomanDubovskyi (approved):

PR #77, opened 2020-12-14
realized methods getAll(), fil() and startWorking()

Alex-Korzun:


nut-pine (changes requested):

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,7 +5,32 @@
...
+    private int id;
+    private String manufacturer;
+
+    public Bulldozer(int id, String manufacturer) {
+        this.id = id;
+        this.manufacturer = manufacturer;
+    }
+
     public Bulldozer() {
+
```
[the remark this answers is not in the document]
Alex-Korzun: removed

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Machine;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Machine> {
+    List<Machine> list = List.of(new Bulldozer(01, "Volvo"),
```
nut-pine, line 8: put this code into your method get()

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Machine> {
+    List<Machine> list = List.of(new Excavator(20, true),
```
nut-pine, line 8: same here

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Machine> {
+    List<Machine> list = List.of(new Truck(20, 1000000),
```
nut-pine, line 8: same here

nut-pine (changes requested):

src/main/java/core/mate/academy/service/BulldozerProducer.java
Alex-Korzun: done

nut-pine (approved):

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Machine;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Machine> {
```
Sofasmile, line 7: I agree with @nut-pine 
why do we parameterize this class with `Machine`? If the name of the class tells us that we will return only `Bulldozers`?

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
...
+
+    @Override
+    public List<Machine> get() {
```
Sofasmile, line 10: the same here and with other classes

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -5,7 +5,7 @@
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
Sofasmile, line 8: if we use only `T` type, we can parameterize this interface with any type of class (It can be Person, String, Object...) Is it valid behavior for `MachineService` interface?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            BulldozerProducer bulldozers = new BulldozerProducer();
```
Sofasmile, line 15: we should operate with abstraction, not an implementation, so let's operate interfaces instead of class here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+            return bulldozers.get();
+        } else if (type == Excavator.class) {
```
Sofasmile, line 17: do we need else block after return statement?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+            ExcavatorProducer excavators = new ExcavatorProducer();
+            return excavators.get();
+        } else {
```
Sofasmile, line 20: let's add this code to the `if` block
and if there is no class type - return an empty list

Sofasmile (changes requested):

src/main/java/core/mate/academy/service/BulldozerProducer.java
Alex-Korzun: Changed parametrization in every class. I had some problems with implementation. But I've resolved it

src/main/java/core/mate/academy/service/BulldozerProducer.java
Alex-Korzun: Done as well

src/main/java/core/mate/academy/service/MachineService.java
Alex-Korzun: Changed to `<T extends Machine>` and made it more specific

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Alex-Korzun: Changed

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Alex-Korzun: removed all `else` blocks

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Alex-Korzun: Done

Sofasmile (approved):
awesome)

PR #83, opened 2020-12-14
add three producers; write 3 methods

vicire:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(Bulldozer.class)) {
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
+            return new ArrayList<>(bulldozerProducer.get());
```
Max-Pochepets, line 18: take out `return` statement out of if-else, make it single in method

Max-Pochepets (changes requested):
nice 👍 one little thing

src/main/java/core/mate/academy/service/MachineServiceImpl.java
vicire: Thanks for review)
Look, I can't take it out coz all of my if statements have a different result to return. That's why I can't make the common `return`. All I can make is to take out the last `else` statement.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+        } else if (type.equals(Excavator.class)) {
+            ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+            return new ArrayList<>(excavatorProducer.get());
+        } else {
+            TruckProducer truckProducer = new TruckProducer();
+            return new ArrayList<>(truckProducer.get());
+        }
+    }
```
Max-Pochepets, line 26:
```suggestion
    public List<Machine> getAll(Class<? extends Machine> type) {
        List<? extends Machine> machines = new ArrayList<>();
        if (type.equals(Bulldozer.class)) {
            machines = new BulldozerProducer().get();
        } else if (type.equals(Excavator.class)) {
            machines = new ExcavatorProducer().get();
        } else {
            machines = new TruckProducer().get();
        }
        return new ArrayList<>(machines);
    }
```

Max-Pochepets (approved):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
vicire: OK, better for me to notice this way :)

Max-Pochepets: 👍

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,7 +5,15 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private String bladeSize;
```
RomanDubovskyi, line 8: how can we access private field? 
add getter/setter
some for other model classes

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, T value);
```
```suggestion
    void fill(List<? super T> machines, T value);
```

RomanDubovskyi (changes requested):

src/main/java/core/mate/academy/model/Bulldozer.java
vicire: Fixed

src/main/java/core/mate/academy/service/MachineService.java
vicire: Fixed

RomanDubovskyi (approved):

PR #84, opened 2020-12-14
PECS

Baldinini:


oroboro-cpu (approved):

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
-     * Return the list of machines.
-     * In the implementation of this method please use your MachineProducer implementations
-     * See that 'Class type' is not parametrized.
-     * Consider to parametrize this Class< PARAMETRIZE_ME > with specific wildcard bounding
-     *
-     * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
-     * @return the list of machines
-     */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
Sofasmile, line 8: If class MachineServiceImpl will be parameterized by Truck. Can we pass Bulldozer.class to the method getAll()? Is it valid behavior?
the same with other methods

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.TruckProducer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
```
Sofasmile, line 16: we should operate with abstraction, not an implementation, so let's operate interfaces instead of class here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+            return new ArrayList<>(bulldozerProducer.get());
+        } else if (type == Excavator.class) {
```
Sofasmile, line 18: do we need else block after return statement?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+            ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+            return new ArrayList<>(excavatorProducer.get());
+        }
+        TruckProducer truckProducer = new TruckProducer();
+        return new ArrayList<>(truckProducer.get());
```
Sofasmile, line 23: let's add this code to the `if` block
and if there is no class type - return an empty list

Sofasmile (changes requested):

Sofasmile (changes requested):
good job)) left one comment

Sofasmile (approved):

PR #85, opened 2020-12-14
implement all classes, realized 3 methods;

mzrkvolodymyr:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+        }
+        if (type == Excavator.class) {
+            ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+            return new ArrayList<>(excavatorProducer.get());
+        }
+        if (type == Truck.class) {
+            TruckProducer truckProducer = new TruckProducer();
+            return new ArrayList<>(truckProducer.get());
+        }
+        return null;
```
RomanDubovskyi, line 29:
```suggestion
        return Collections.EMPTY_LIST;
```
it's better to return empty list instead to avoid NPE

RomanDubovskyi (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
```
RomanDubovskyi, line 19:
```suggestion
            MachineProducer bulldozerProducer = new BulldozerProducer();
```
better to operate with abstraction

RomanDubovskyi (approved):

PR #86, opened 2020-12-14
Task completed

Vedroid:


AndreiYazadzhi (approved):

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
-
+    void fill(List<? super Machine> machines, T value);
```
```suggestion
    void fill(List<? super T> machines, T value);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,44 @@
...
+            List<? extends Machine> machines = bulldozerProducer.get();
+            return new ArrayList<>(machines);
+        }
+        if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        if (type == Truck.class) {
+            return new ArrayList<>(new TruckProducer().get());
+        }
+        return null;
```
RomanDubovskyi, line 28: FYI: 
```suggestion
        return Collections.EMPTY_LIST;
```

RomanDubovskyi (approved):

PR #97, opened 2020-12-15
all classes done

vlykhvar:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,6 +5,7 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+
```
gigbat, line 8: you need to read file with typical mistakes. You need to add some fields and initialize those using super(...);

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,12 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Machine;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<? extends Machine> get() {
+        return List.of(new Bulldozer(),new Bulldozer(), new Bulldozer());
```
gigbat, line 10: when you added some fields you need to give those as parameters for some objects

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,12 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
```
gigbat, line 7: read previous

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
...
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends Machine> get();
```
gigbat, line 11: `List<T> get();` it's correct variant

gigbat: because your interface is a producer, already

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+import java.util.List;
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(Bulldozer.class)) {
```
gigbat, line 15: it's up to you, but you can compare using ==. Equals need to compare values, but == need to compare links

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+            return (List<Machine>) new BulldozerProducer().get();
+        }
+        if (type.equals(Excavator.class)) {
+            return (List<Machine>) new ExcavatorProducer().get();
+        }
+        if (type.equals(Truck.class)) {
+            return (List<Machine>) new TruckProducer().get();
+        }
+        return null;
```
gigbat, line 24: Sofia will say you that you return an empty list using the method of class Collections

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,12 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+    @Override
+    public List<? extends Machine> get() {
```
gigbat, line 9: we should be using a specified class instead of upper bounding. e.g. <Truck> for TruckProducer and etc.

gigbat (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
vlykhvar: changed

src/main/java/core/mate/academy/service/MachineServiceImpl.java
vlykhvar: ok. Changed

gigbat (approved):

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,12 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Machine;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<? super Machine> get() {
```
Sofasmile, line 9: can we return any type of machine here, if the interface is parameterized with Bulldozer?
let's replace `? super Machine` with Bulldozer too
the same with other producers

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,8 @@
...
 import java.util.List;
 
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<T extends Machine> {
+    List<?> get();
```
Sofasmile, line 7: why do we use `?` here instead of `T`?

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
Sofasmile, line 9: if we use only `T` type, we can parameterize this interface with any type of class (It can be Person, String, Object...) Is it valid behavior for `MachineService` interface?

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, T value);
```
Sofasmile, line 30: Why do we use Machine here instead of T type?

Sofasmile (changes requested):

src/main/java/core/mate/academy/service/MachineProducer.java
vlykhvar: couse after i changed to T nothig works

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,7 +5,14 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private String name;
```
Sofasmile, line 8: let's add getters for fields

Sofasmile (approved):

PR #100, opened 2020-12-15
add TruckProducer, ExcavatorProducer and BulldozerProducer

IwalkInSilence:
parametrize MachineProducer, MachineService, MachineServiceImpl and others
add private fields, getters and setters and constructors

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineService.java
+++ b/src/main/java/core/mate/academy/service/MachineService.java
@@ -1,12 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
-
+public interface MachineService<T extends Machine> {
     /**
      * Return the list of machines.
      * In the implementation of this method please use your MachineProducer implementations
@@ -16,7 +16,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class type);
 
     /**
      * Fill the machines list with passed value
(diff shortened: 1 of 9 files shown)


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +26,13 @@
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, Object value);
```
Artificial-Friend, line 29: Consider changing `Object` to exact parameter you should get

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +16,7 @@
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class type);
```
Artificial-Friend, line 19: It is better to parametrize `Class` so there will be only a class that we require with no mistakes

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import core.mate.academy.model.TruckProducer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class type) {
+        List<? extends Machine> resultList = new ArrayList<>();
```
Artificial-Friend, line 16: We can use `return` in cycle therefore if first condition is met we would not need to check the other two. Also  it is better to return `Collections.EMPTY_LIST` so we will not have NPE in future.

Artificial-Friend (changes requested):

Artificial-Friend (approved):

RomanDubovskyi (approved):

PR #108, opened 2021-04-06
Wildcart Machines Solution

SakVladyslav:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService {
```
RomanDubovskyi, line 13:
```suggestion
public class MachineServiceImpl implements MachineService<Machine> {
```

src/test/java/core/mate/academy/service/MachineServiceTest.java
```diff
@@ -14,7 +14,7 @@
 import java.util.Random;
 
 public class MachineServiceTest {
-    private static MachineService<Machine> machineService;
+    private static MachineService machineService;
```
RomanDubovskyi, line 17: don't change test class

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<Machine> machines = new ArrayList<>();
+
+        if (type == Bulldozer.class) {
+            machines.add(new Bulldozer());
```
RomanDubovskyi, line 19: we should call correct MachineProducer according to the task description
image

RomanDubovskyi (changes requested):

src/main/java/core/mate/academy/model/BulldozerProducer.java
```diff
@@ -0,0 +1,17 @@
+package core.mate.academy.model;
+
+import core.mate.academy.service.MachineProducer;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
```
RomanDubovskyi, line 6:
```suggestion
public class BulldozerProducer implements MachineProducer<Bulldozer> {
```
it's better not to use raw parameters here. This way you will be able to return only Bulldozers from get method

src/main/java/core/mate/academy/model/BulldozerProducer.java
```diff
@@ -0,0 +1,17 @@
...
+    private final Bulldozer bulldozer1 = new Bulldozer();
+    private final Bulldozer bulldozer2 = new Bulldozer();
+    private final Bulldozer bulldozer3 = new Bulldozer();
+
+    private final List<Bulldozer> bulldozers = List.of(bulldozer1, bulldozer2, bulldozer3);
+
+    @Override
+    public List<Bulldozer> get() {
+        return bulldozers;
```
RomanDubovskyi, line 15: now this class will always return the same 3 bulldozers. It's better to make them as local variables and always return new list with new bulldozers. That's default behaviour for producer

src/main/java/core/mate/academy/model/ExcavatorProducer.java
```diff
@@ -0,0 +1,17 @@
+package core.mate.academy.model;
+
+import core.mate.academy.service.MachineProducer;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer {
```
RomanDubovskyi, line 6: same comments as for BulldozerProducer

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #117, opened 2021-04-12
accepted task

kupora3:


src/main/java/core/mate/academy/model/TruckProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.model;
+
+import core.mate.academy.service.MachineProducer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TruckProducer extends Truck implements MachineProducer {
+
+    @Override
+    public List<? super Truck> get() {
```
IhorRome, line 10: it looks like we can add to result list some object of class Machine and Object (same for two other implemantations)
```suggestion
    public List<Truck> get() {
```

src/main/java/core/mate/academy/model/TruckProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.model;
+
+import core.mate.academy.service.MachineProducer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TruckProducer extends Truck implements MachineProducer {
```
IhorRome, line 7:
```suggestion
public class TruckProducer implements MachineProducer<Truck> {
```
we don't need to extend class Truck here, but have to paramatrize imlemented interface
same for two other classes

IhorRome (changes requested):

src/main/java/core/mate/academy/model/BulldozerProducer.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.List;
+
+public class BulldozerProducer extends Bulldozer implements MachineProducer {
+
+    @Override
+    public List<? super Bulldozer> get() {
+        List<Bulldozer> bulldozer = new ArrayList<>();
+        bulldozer.add(new Bulldozer());
+        bulldozer.add(new Bulldozer(10,12));
+        return bulldozer;
```
IhorRome, line 14: not a mistake, I guess, but you dont have to create new ArrayList here
```suggestion
        return List.of(new Bulldozer(...), new Bulldozer());
```

kupora3: I think it's the same

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,9 @@
...
 import java.util.List;
 
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<T extends Machine> {
+
```
freedmanMain, line 7: extra empty line =))))

freedmanMain (approved):

PR #122, opened 2021-04-18
add solution

KononenkoKonstantin:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,49 @@
...
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List getAll(Class type) {
+        List<? extends Machine> listOfMachines;
+        switch (type.getSimpleName()) {
+            case "Truck": {
+                listOfMachines = new TruckProducer().get();
```
RomanDubovskyi, line 17: уwe can return here list right away, and it's better to create instances of producers not to generate new objects per each method call

RomanDubovskyi (approved):
nice

src/main/java/core/mate/academy/service/MachineServiceImpl.java
KononenkoKonstantin: I've done it, If I understood right

PR #136, opened 2021-07-16
jv-pecs-hw

SHTORMICH:
Add 3 class (BulldozerProducer, ExcavatorProducer, TruckProducer) and implement method get. Parameterize the methods in class MachineService. Than realize methods in MachineServiceImpl which implementing MachineService.


RomanDubovskyi (approved):
nice work 🔥

PR #149, opened 2021-08-30
PECS was done

Yelinskyi:


kovalskiy7777 (approved):
Well done!!!!

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,14 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
```
MaliukDaria, line 9: lets add few of custom buldozers with some parameters to the list (same for all classes)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(Bulldozer.class)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if ((type.equals(Excavator.class))) {
```
MaliukDaria, line 18: do we need `else` after `return`?

MaliukDaria (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Yelinskyi: It isn't necessarily.
I noticed this mistakes in someone else's code when i reviewed, but did the same.

I corrected it.

src/main/java/core/mate/academy/service/BulldozerProducer.java
Yelinskyi: I added some machines in each class.

PR #165, opened 2021-09-25
made classes and methods

dogovor:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        List<? extends Machine> machines = new ArrayList<>();
+        if (type == Bulldozer.class) {
+            machines = bulldozerProducer.get();
+        } else if (type == Excavator.class) {
+            machines = excavatorProducer.get();
+        } else if (type == Truck.class) {
+            machines = truckProducer.get();
+        }
+        return new ArrayList<>(machines);
+    }
```
Black3942, line 26: we create 3 object but return just a one.
did we realy need to create other objects?

Black3942 (changes requested):

Black3942 (approved):

Resci (approved):

PR #167, opened 2021-10-02
Created BulldozerProducer, ExcavatorProducer, TruckProducer classes and implemented their methods. Implemented methods in MachineServiceImpl.

MiaHrynda:


sich-mykhailo (approved):

src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -4,6 +4,12 @@
     private String name;
     private String color;
 
+    public Machine() {
+    }
```
Resci, line 8: Where do you use this constructor?

src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -4,6 +4,12 @@
...
+
+    public Machine(String name, String color) {
+    }
```
Resci, line 11: Why do you ignore `name` and `color` from constructor parameters?

Resci (approved):

PR #171, opened 2021-10-14
Implement PECS, practise.

osa240:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private List<Bulldozer> bulldozers = new ArrayList<>();
+
+    {
+        Bulldozer bulldozer = new Bulldozer();
+        bulldozer.setName("Bulldozer");
+        bulldozer.setColor("White");
+        bulldozer.setWight(50);
+        bulldozers.add(bulldozer);
```
Yelinskyi, line 15: I think It would be better if you  added at list 2 instance of Bulldozer.
And the same in Excavator and Truck

PR #206, opened 2022-01-05
Hw pecs main

maxkmaxk:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,50 @@
...
+        if (Bulldozer.class.equals(type)) {
+            return new BulldozerProducer();
+        }
+        if (Excavator.class.equals(type)) {
+            return new ExcavatorProducer();
+        }
+        if (Truck.class.equals(type)) {
+            return new TruckProducer();
+        }
+        return null;
```
MaliukDaria, line 48: It's bad practice to return null, lets throw some exception instead

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,50 @@
...
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer producer = getProducer(type);
+        if (producer == null) {
+            return new ArrayList<>();
+        }
+        List<? extends Machine> retVal = producer.get();
```
MaliukDaria, line 20: retVal - not informative exception message

MaliukDaria (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
maxkmaxk: I agree, moreover initially this function thrown IllegalArgumentException. But, one test (getAll_invalidType()) faulted. So, i checked [https://mate-academy.github.io/jv-program-common-mistakes/java-core/pecs/pecs.html] and find out there : 

> Don’t return null or specific type List in getAll() when no producer has been found.
Returning null is considered to be a bad practice: a user of your method needs to write additional null checks when using it. In other cases NPE might appear. Returning List of a specific type, like List<Truck> when no producer found is a bad practice as well, because it’s not an obvious behaviour, instead it’s better to return empty List.

That's why method returns empty List

MaliukDaria: Agree, lets return empty list

MaliukDaria (changes requested):
I don't see changes from previous review

src/main/java/core/mate/academy/service/MachineServiceImpl.java
maxkmaxk: Sorry, i don't understand what it mean - there is nothing about exception in this code. retVal using only as source for method result : 
`public class MachineServiceImpl implements MachineService<Machine> {
    @Override
        MachineProducer producer = getProducer(type);
        if (producer == null) {
            return new ArrayList<>();
        }
        List<? extends Machine> retVal = producer.get();
        return new ArrayList<>(retVal);
    }`
What's wrong with it ?

maxkmaxk:
> 

I don't know what to change. Looks like some misunderstanding... could you, please, explain me exactly what's wrong with my code

src/main/java/core/mate/academy/service/MachineServiceImpl.java
MaliukDaria: retVal - не информативное название переменной. Плохая практика использовать сокращения в названиях, давай ее переименуем так, что б сразу было понятно, что за значение лежит в этой переменной. Например `machines`

MaliukDaria, line 20:
```suggestion
        List<? extends Machine> machines = producer.get();
```

MaliukDaria (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
maxkmaxk: Спасибо! Теперь понял. Фраза "not informative exception message" смутила, опечатка наверное.

RomanDubovskyi (approved):

PR #207, opened 2022-01-05
Pecs work completed.

StanislavSonich:


src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -19,4 +15,8 @@ public String getColor() {
     public void setColor(String color) {
         this.color = color;
     }
+
+    public abstract void bucketVolume(double v);
+
+    public abstract void setLoadCapacity(int i);
```
MaliukDaria, line 21: Why do we need this methods? What are they doing?

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> listOfBulldozers = new ArrayList<>();
+        for (int i = 0; i < 3; i++) {
+            listOfBulldozers.add(new Bulldozer());
+        }
+        return listOfBulldozers;
```
MaliukDaria, line 14: Why do you create list of trucks with some random parameter values and list of Bulldozers and Excavators with no parameter values? Lets add some parameter values to Bulldozer and Excavator lists

src/main/java/core/mate/academy/model/TruckProduser.java
```diff
@@ -0,0 +1,33 @@
+package core.mate.academy.model;
+
+import core.mate.academy.service.MachineProducer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TruckProduser implements MachineProducer<Truck> {
```
MaliukDaria, line 7: Do we need two TruckProdusers? (TruckProduser, TruckProdcer)

MaliukDaria (changes requested):

RomanDubovskyi (approved):

PR #219, opened 2022-01-29
Hw solution

Sasha-Bachynskyi:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+        List<Truck> trucks = new TruckProducer().get();
+        if (type.equals(Bulldozer.class)) {
+            return bulldozers;
+        }
+        if (type.equals(Excavator.class)) {
+            return excavators;
+        }
+        if (type.equals(Truck.class)) {
+            return trucks;
+        }
```
LinkClink, line 24: We can use this  model - "else if"

LinkClink (approved):
LGTM

src/main/java/core/mate/academy/service/MachineServiceImpl.java
AlexRozv: agree, at the moment you're always creating 3 lists of machines, let's use if-else to create and return a specific list

AlexRozv (changes requested):

AlexRozv (approved):
well done 👍

PR #238, opened 2022-02-23
Hw jv pecs

VitaliiBozadzhy88:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,11 +5,44 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private String ladle;// ковш
```
RomanDubovskyi, line 8: Use comments, where you want to explain why this or that decision was made if it's not obvious. No need to leave translates🙂

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
+    @Override
+    public List<Bulldozer> get() {
+        Bulldozer bulldozer1 = new Bulldozer("has a ladle", "has caterpillars");
+        Bulldozer bulldozer2 = new Bulldozer("has not ladle", "has caterpillars");
+        Bulldozer bulldozer3 = new Bulldozer("has a ladle", "has not caterpillars");
+
+        return List.of(bulldozer1, bulldozer2,bulldozer3);
```
RomanDubovskyi, line 13:
```suggestion
        return List.of(bulldozer1, bulldozer2, bulldozer3);
```

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,14 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer {
```
RomanDubovskyi, line 6: Now, this class is not parameterized (we didn't declare what type is playing the role of <T> from MachineProducer. Let's fix that

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
```
RomanDubovskyi, line 5: Now we can use any type for T in our impls. But it does not make sense to have String or Object instead of T. As a solution we can restrict types using T extends .... in here

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -5,7 +5,7 @@
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
RomanDubovskyi, line 8: same here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,35 @@
...
 
-/**
- * Your implementation of MachineService.
- */
-public class MachineServiceImpl {
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
```
RomanDubovskyi, line 7: do the same way with producers when you declare those classes

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,35 @@
...
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        switch (type.getSimpleName()) {
```
RomanDubovskyi, line 11: GJ 👍

RomanDubovskyi (changes requested):

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,16 @@
...
-    /**
-     * Fill the machines list with passed value
-     * Replace the Object with parametrized value.
-     * This method should be able to work well with any type of machine passed as 'value'
-     *
-     * @param machines - list of machines to be filled with value
-     * @param value    - any object of machine sub class
-     */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, Machine value);
```
RomanDubovskyi, line 13:
```suggestion
    void fill(List<? super T> machines, T value);
```

RomanDubovskyi (approved):

PR #249, opened 2022-03-24
my solution is ready

DenysShl:


src/main/java/core/mate/academy/MachineApp.java
```diff
@@ -0,0 +1,21 @@
+package core.mate.academy;
+
+import core.mate.academy.model.Machine;
+import core.mate.academy.service.BulldozerProducer;
+import core.mate.academy.service.ExcavatorProducer;
+import core.mate.academy.service.TruckProducer;
+import java.util.List;
+
+public class MachineApp {
```
MaliukDaria, line 9: let remove this class from PR

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,11 +5,32 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private String markBulldozer;
```
MaliukDaria, line 8:
```suggestion
    private String mark;
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #279, opened 2022-05-24
implemented task

gdpbalt:
- add model classes: Bulldozer, Excavator, Truck;
- add classes for generate list of models: MachineProducer, BulldozerProducer, ExcavatorProducer, TruckProducer;
- change interface MachineService;
- add implementation of MachineService in class MachineServiceImpl.


liliia-ponomarenko (approved):
good job 👍

PR #281, opened 2022-05-26
Initial solution commit

Oleksii-Mospan:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,11 +5,27 @@
...
 public class Bulldozer extends Machine {
+    private String dumpShape;
+
     public Bulldozer() {
     }
 
+    public Bulldozer(String name, String color, String dumpShape) {
+        this.dumpShape = dumpShape;
+        this.setColor(color);
+        this.setName(name);
```
Nick97-git, line 16: don't use setters in constructor, to initialize these fields, use constructor of a parent, for such cases + I don't see such setters here in this class, the same for other figures

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
Nick97-git, line 9:
```suggestion
public interface MachineService<T extends Machine> {
```

Nick97-git (changes requested):

src/main/java/core/mate/academy/service/MachineService.java
Oleksii-Mospan: When i try this, I catch a lot of mistakes and don`t know how to fix it all ((.

Nick97-git: tag me in your channel, about this problem and prepare screen with errors

Nick97-git (changes requested):

Nick97-git (approved):

PR #287, opened 2022-05-30
Implemented BulldozerProducer, ExcavatorProducer, TruckProducer.

ChumakAndrey:


src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -5,9 +5,25 @@
...
  */
 public class Truck extends Machine {
+    private int loadCapacity;
+
     public Truck() {
     }
 
+    public Truck(String name, String color, int loadCapacity) {
+        setName(name);
+        setColor(color);
```
KostyaNikiforov, line 15: Let's use a parental constructor instead of setters

KostyaNikiforov (changes requested):

PR #292, opened 2022-06-01
create class BulldozerProducer, ExcavatorProducer, TruckProducer and …

IDMFokin:
…implemented interface MachineProduser, MachineServicelImpl

### DIFF
--- a/src/main/java/core/mate/academy/model/Bulldozer.java
+++ b/src/main/java/core/mate/academy/model/Bulldozer.java
@@ -4,8 +4,36 @@ package core.mate.academy.model;
  * Add some custom fields that could be only in Bulldozer
  * Do not remove no-args constructor
  */
+
 public class Bulldozer extends Machine {
+
+    private int shieldWidth;
+    private int shieldHeight;
+
     public Bulldozer() {
+    }
+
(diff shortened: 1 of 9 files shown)


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -4,10 +4,38 @@
...
+
+    private int shieldWidth;
+    private int shieldHeight;
+
     public Bulldozer() {
     }
 
+    public Bulldozer(String name, String color, int shieldWidth, int shieldHeight) {
+        super.setName(name);
+        super.setColor(color);
```
MaliukDaria, line 18: do we need `super` keyword here?

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
```
MaliukDaria, line 5:
```suggestion
public interface MachineProducer<T extends Machine> {
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #294, opened 2022-06-02
methods were implemented

mykola8426:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        } else if (Excavator.class.equals(type)) {
+            ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+            list = excavatorProducer.get();
+        } else if (Truck.class.equals(type)) {
+            TruckProducer truckProducer = new TruckProducer();
+            list = truckProducer.get();
+        } else {
+            return new ArrayList<>();
+        }
+        return new ArrayList<>(list);
```
oleksii-kordysh, line 27: The logic is correct, but what do you think, if we just assign smth for list value (for ex.: **list = new SmthProducer()**) in if-statement's body, and after all the checks return list.get() ? It will short your code and you could not create new ArrayList in the end.

oleksii-kordysh (changes requested):
Hi, got some ideas on how to improve some code.

oleksii-kordysh (approved):

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

PR #296, opened 2022-06-05
implemented PECS task

DimaPodakin:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class<? extends Machine> type);
```
liliia-ponomarenko, line 20:
```suggestion
    List<T> getAll(Class<? extends T> type);
```
Let's use abstraction inside the interface

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

liliia-ponomarenko (approved):
LGTM 🔥

PR #316, opened 2022-07-24
hw-jv-pecs

serhii-davydov:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private List<Bulldozer> list;
+
+    {
+        list = new ArrayList<>();
+        list.add(new Bulldozer("Cat 9DT", "yellow"));
+        list.add(new Bulldozer("John Deere 1050K", "yellow"));
+    }
```
MaliukDaria, line 14: I think it's better to make this variable local (same it other classes)

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class<? extends Machine> type);
```
MaliukDaria, line 20:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

MaliukDaria (changes requested):
Lets add some type-specific fields to all Mathine classes.

MaliukDaria (approved):

PR #328, opened 2022-08-09
Described solution was added.

DioStesso1:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
```
Sofasmile, line 5: now we can do the following:
```
class StringProducer implement MachineProducer<String>
```
is it correct?

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -5,7 +5,7 @@
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
Sofasmile, line 8: same here
can MachineService work with String, Integer or People or other non-machine classes?

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +16,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List getAll(Class<? extends T> type);
```
Sofasmile, line 19: we shouldn't remove parametrization from List

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List getAll(Class<? extends Machine> type) {
+        if (Bulldozer.class.equals(type)) {
+            return new ArrayList(new BuldozerProducer().get());
+        } else if (Excavator.class.equals(type)) {
```
Sofasmile, line 19: after return we can remove else block

Sofasmile (changes requested):

PR #329, opened 2022-08-09
implemented machine service

dmytrovb02:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+            return new ArrayList<>(new TruckProducer().get());
+        }
+        if (type.equals(Bulldozer.class)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        }
+        if (type.equals(Excavator.class)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else {
+            return new ArrayList<>();
+        }
```
MaliukDaria, line 26: could we initialize MachineProducers in MachineServiceImpl class, not to create new instance of  TruckProducer, BulldozerProducer, ... every time ve call `getAll()` method?

MaliukDaria (approved):

PR #350, opened 2022-09-05
HW done

oleksandr-kravchenko:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        }
+        if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        if (type == Truck.class) {
+            return new ArrayList<>(new TruckProducer().get());
+        }
+        return new ArrayList<>();
```
MaliukDaria, line 25: It's better to create all MachineProducers objects(new BulldozerProducer(), new ExcavatorProducer(), ...) as class level variables, not to create instances every time we call getAll() method

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        int size = machines.size();
```
MaliukDaria, line 30: do we need this variable, can we just call machines.size() where needed?

MaliukDaria (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
oleksandr-kravchenko: > It's better to create all MachineProducers objects(new BulldozerProducer(), new ExcavatorProducer(), ...) as class level variables, not to create instances every time we call getAll() method

didn't understand what to do

PR #352, opened 2022-09-05
pecs implemented

Litvynets:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,49 @@
...
+                MachineProducer<Excavator> excavatorProducer = new ExcavatorProducer();
+                return new ArrayList<>(excavatorProducer.get());
+            }
+            if (type == Truck.class) {
+                MachineProducer<Truck> truckProducer = new TruckProducer();
+                return new ArrayList<>(truckProducer.get());
+            }
+        }
+        return new ArrayList<>();
+    }
```
aaananas, line 31:
```suggestion
public class MachineServiceImpl implements MachineService<Machine> {
    private final MachineProducer<Bulldozer> bulldozerProducer = new BulldozerProducer;
    private final MachineProducer<Excavator> excavatorProducer = new ExcavatorProducer();
    private final MachineProducer<Truck> truckProducer = new TruckProducer();
    
    @Override
    public List<Machine> getAll(Class<? extends Machine> type) {
        if (type != null) {
            if (type == Bulldozer.class) {
                return new ArrayList<>(bulldozeeProducer.get());
            }
            if (type == Excavator.class) {
                return new ArrayList<>(excavatorProducer.get());
            }
            if (type == Truck.class) {
                return new ArrayList<>(truckProducer.get());
            }
        }
        return new ArrayList<>();
    }
```
you can create producers just once

aaananas (approved):

PR #369, opened 2022-10-03
Machinery service implementation

shmex-j:


src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,19 @@
...
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+    private final List<Truck> truckList;
+
+    public TruckProducer() {
+        truckList = new ArrayList<>();
+        truckList.add(new Truck(15d));
```
Bodddd, line 12: it can be magic numbers, but not sure

Bodddd (changes requested):

Bodddd (approved):
Good work🔥

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,51 @@
...
+        if (type == Bulldozer.class) {
+            machines = bulldozerProducer.get();
+        }
+        if (type == Excavator.class) {
+            machines = excavatorProducer.get();
+        }
+        if (type == Truck.class) {
+            machines = truckProducer.get();
+        }
+        return new ArrayList<>(machines);
```
KostyaNikiforov, line 37: Your solution looks good but this code must be a good example of polymorphism.
Let's create an abstract variable of a type MachineProducer and initialize it with an applicable producer implementation. Then just call the method get one time.

------
(optional) Also, you can try to use a hashMap instead of if-blocks. It is a really good practice.

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
Good job!:fire:

PR #372, opened 2022-10-06
implementation of these interfaces - MachineProducer and MachineService

Bodddd:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        MachineProducer<? extends Machine> machineProducer;
+        if (Bulldozer.class.equals(type)) {
+            machineProducer = new BulldozerProducer();
+        } else if (Truck.class.equals(type)) {
+            machineProducer = new TruckProducer();
+        } else if (Excavator.class.equals(type)) {
+            machineProducer = new ExcavatorProducer();
+        } else {
+            return new ArrayList<>();
+        }
```
shmex-j, line 25: You can use == instead equals

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+            machineProducer = new TruckProducer();
+        }
+        if (Excavator.class == type) {
+            machineProducer = new ExcavatorProducer();
+        }
+        if (machineProducer == null) {
+            return new ArrayList<>();
+        }
+        List<? extends Machine> machines = machineProducer.get();
+        return new ArrayList<>(machines);
```
KostyaNikiforov, line 30:
```suggestion
        return machineProducer == null ? new ArrayList<>(machineProducer.get()) : Collection.emptyList();
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
🔥

PR #374, opened 2022-10-08
implemented MachineService and methods

andrey156:


zachar-chorny (changes requested):
Screen Shot 2022-10-08 at 7 14 51 PM
Fix it

PR #376, opened 2022-10-09
MachineService implemented

hpavlo:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
...
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends Machine> get();
```
alex-anofriev, line 11: I think you can use here just T because in interface you already wrote that T extends Machine

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, Machine value);
```
alex-anofriev, line 30: Here also I guess Machine can be replaced with T as in interface you wrote that T extends Machine

alex-anofriev (approved):

ChumakAn (approved):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class type) {
+        if (Bulldozer.class.equals(type)) {
+            return new LinkedList<>(new BulldozerProducer().get());
```
ChumakAn, line 17: why do you use the LinkedList?

ChumakAn (changes requested):
Well done, if you want to make it better, than ->
image
it's better to add some logic to the hierarchy of classes, also you can use super() in child classes to inherit the fields and make this class structure better-looking
Also, what's the reason to use the LinkedList?

ChumakAn (approved):

PR #381, opened 2022-10-10
implemented MachineService, MachineProducer classes, added RandomValu…

GrishaSeniv:
implemented MachineService, MachineProducer classes, added RandomValue in util package

### DIFF
--- a/src/main/java/core/mate/academy/service/ExcavatorProducer.java
+++ b/src/main/java/core/mate/academy/service/ExcavatorProducer.java
@@ -1,0 +1,51 @@
+package core.mate.academy.service;
+
+import core.mate.academy.constant.Color;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.util.RandomValue;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    private RandomValue randomValue;
+    private Excavator greenExcavator;
+    private Excavator yellowExcavator;
+    private Excavator redExcavator;
+    private List<Excavator> listExcavator;
+
+    public ExcavatorProducer() {
+        randomValue = new RandomValue();
+        this.listExcavator = new ArrayList<>();
+    }
+
+    @Override
+    public List<Excavator> get() {
+        init();
+        listExcavator.add(greenExcavator);
+        listExcavator.add(yellowExcavator);
+        listExcavator.add(redExcavator);
+        return listExcavator;
+    }
+
+    @Override
+    public void init() {
+        greenExcavator = new Excavator();
+        greenExcavator.setName(Excavator.class.getName());
+        greenExcavator.setColor(Color.GREEN.name());
+        greenExcavator.setArmLength(randomValue.getRandomValue());
+        greenExcavator.setBoomLength(randomValue.getRandomValue());
+        greenExcavator.setTurningRadius(randomValue.getRandomValue());
+        yellowExcavator = new Excavator();
+        yellowExcavator.setName(Excavator.class.getName());
+        yellowExcavator.setColor(Color.YELLOW.name());
+        yellowExcavator.setArmLength(randomValue.getRandomValue());
+        yellowExcavator.setBoomLength(randomValue.getRandomValue());
+        yellowExcavator.setTurningRadius(randomValue.getRandomValue());
+        redExcavator = new Excavator();
+        redExcavator.setName(Excavator.class.getName());
+        redExcavator.setColor(Color.RED.name());
+        redExcavator.setArmLength(randomValue.getRandomValue());
+        redExcavator.setBoomLength(randomValue.getRandomValue());
+        redExcavator.setTurningRadius(randomValue.getRandomValue());
+    }
+}
--- a/src/main/java/core/mate/academy/service/MachineServiceImpl.java
+++ b/src/main/java/core/mate/academy/service/MachineServiceImpl.java
@@ -1,7 +1,41 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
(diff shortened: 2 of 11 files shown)


src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,51 @@
...
+import core.mate.academy.model.Excavator;
+import core.mate.academy.util.RandomValue;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    private RandomValue randomValue;
+    private Excavator greenExcavator;
+    private Excavator yellowExcavator;
+    private Excavator redExcavator;
```
CyberDogFK, line 13: I think this field are not related to the state of the object. So, much better decision it`s create method, like madeDefaultYellowTruk(), madeDefaultRedTruck(), etc. And make this method`s return created objects, and add this object`s to list from those method`s, not from fileds

CyberDogFK: And this help you do make init() method less)

CyberDogFK (changes requested):

CyberDogFK (approved):
Good job!)

src/main/java/core/mate/academy/util/RandomValue.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.util;
+
+import java.util.Random;
+
+public class RandomValue {
```
KostyaNikiforov, line 5:
```suggestion
public class RandomValueGenerator {
```
or 
```suggestion
public class RandomValueService {
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
```
KostyaNikiforov, line 13:
```suggestion
public class MachineServiceImpl implements MachineService<Machine> {
```

KostyaNikiforov: Here we work with implementation, we don't need to use abstract types anymore

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+    @Override
+    public List<T> getAll(Class<? extends T> type) {
```
KostyaNikiforov, line 15:
```suggestion
    public List<Machine> getAll(Class<? extends Machine> type) {
```
Fix it everywhere

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+        if (type == Bulldozer.class) {
+            return (List<T>) new BulldozerProducer().get();
+        }
+        if (type == Excavator.class) {
+            return (List<T>) new ExcavatorProducer().get();
+        }
+        if (type == Truck.class) {
+            return (List<T>) new TruckProducer().get();
+        }
+        return new ArrayList<>();
```
KostyaNikiforov, line 25: This code must be a good example of a polymorphism.
Let's create an abstract variable of a type MachineProducer and initialize it with an applicable producater implementation. Then just call the method get one time.

------
(optional) Also, you can try to use a hashMap instead of if-blocks. It is a really good practice.

KostyaNikiforov (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,49 @@
...
+    }
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer<? extends Machine> machineProducer;
+        machineProducer = machineProducerMap.get(type);
+        if (machineProducer == null) {
+            return new ArrayList<>();
+        }
+        return new ArrayList<>(machineProducer.get());
```
KostyaNikiforov, line 33:
```suggestion
        return machineProducer == null ? new ArrayList<>(machineProducer.get()) : new ArrayList<>();
```
or even
```suggestion
        return machineProducer == null ? new ArrayList<>(machineProducer.get()) : Collection.emptyList();
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
GrishaSeniv: > This code must be a good example of polymorphism. Let's create an abstract variable of a type MachineProducer and initialize it with an applicable producer implementation. Then just call the method get one time.
> 
> (optional) Also, you can try to use a hashMap instead of if-blocks. It is a really good practice.

Good idea

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,49 @@
...
+        machineProducerMap = new HashMap<>();
+        machineProducerMap.put(Bulldozer.class, new BulldozerProducer());
+        machineProducerMap.put(Excavator.class, new ExcavatorProducer());
+        machineProducerMap.put(Truck.class, new TruckProducer());
+    }
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer<? extends Machine> machineProducer;
+        machineProducer = machineProducerMap.get(type);
```
KostyaNikiforov, line 29:
```suggestion
        MachineProducer<? extends Machine> machineProducer = machineProducerMap.get(type);
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
Good job!:fire:

PR #383, opened 2022-10-12
homework has been done

AndriiMatveev:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,12 +1,7 @@
...
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
-    public Bulldozer() {
-    }
+    private int size = 0;
 
     @Override
     public void doWork() {
```
arthur4655, line 7: /**
 */

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,12 +1,7 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
-    public Bulldozer() {
-    }
+    private int size = 0;
```
arthur4655, line 4: Also size = 0 is redundant

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+            machines = new BulldozerProducer().get();
+        }
+        if (type == Excavator.class) {
+            machines = new ExcavatorProducer().get();
+        }
+        if (type == Truck.class) {
+            machines = new TruckProducer().get();
+        }
+        return machines;
+    }
```
arthur4655, line 24: We can use return in if-else body. After that creating empty list will be redundant, as we can return list of machines from return statement, and empty list of machines in default case

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+        for (int i = 0; i < machines.size(); i++) {
+            machines.set(i,value);
+        }
+    }
+
+    @Override
+    public void startWorking(List<? extends T> machines) {
+        if (machines == null) {
+            return;
+        }
```
arthur4655, line 40: This is redundant, if list of machines will be null, cycle  just won't start and exit from method

arthur4655 (changes requested):

src/main/java/core/mate/academy/model/Bulldozer.java
arthur4655: image

src/main/java/core/mate/academy/model/Bulldozer.java
arthur4655: image

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+        switch (type.getSimpleName()) {
+            case "Bulldozer":
+                return machines = new BulldozerProducer().get();
+            case "Excavator":
+                return machines = new ExcavatorProducer().get();
+            case "Truck":
+                return machines = new TruckProducer().get();
+            default:
+                return machines;
+        }
```
arthur4655, line 20: I mean we don't  need line List of machines = new ArrayList<>():
We can write for example "return new BulldozerProducer().get()" so 10 line will be redundant.

arthur4655 (changes requested):

arthur4655 (approved):

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,13 +1,19 @@
...
- */
 public class Bulldozer extends Machine {
+    private int size;
+
     public Bulldozer() {
     }
 
+    public Bulldozer(int size) {
+        this.size = size;
+    }
```
ChumakAn, line 11: what about inheriting the fields from Machine and passing them in constructor?

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -1,13 +1,19 @@
...
- * Do not remove no-args constructor
- */
 public class Excavator extends Machine {
+    private int weight;
+
     public Excavator() {
     }
 
+    public Excavator(int weight) {
+        this.weight = weight;
```
ChumakAn, line 10: what about inheriting the fields from Machine and passing them in constructor?

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -1,13 +1,19 @@
...
- * Do not remove no-field constructor
- */
 public class Truck extends Machine {
+    private boolean isLoaded;
+
     public Truck() {
     }
 
+    public Truck(boolean isLoaded) {
+        this.isLoaded = isLoaded;
```
ChumakAn, line 10: what about inheriting the fields from Machine and passing them in constructor?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
 
-/**
- * Your implementation of MachineService.
- */
-public class MachineServiceImpl {
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
```
ChumakAn, line 7: reconsider using the <>, <T extends Machine> - do you need this?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
+    @Override
+    public List<Machine> getAll(Class<? extends T> type) {
+        switch (type.getSimpleName()) {
+            case "Bulldozer":
```
ChumakAn, line 11: Optional, but what about comparing the class names, not simpleNames?

ChumakAn (changes requested):
Good, but let's make it better

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -1,13 +1,21 @@
...
- */
 public class Excavator extends Machine {
+    private int weight;
+
     public Excavator() {
     }
 
+    public Excavator(String name, String colour, int weight) {
+        this.setName(name);
+        this.setColor(colour);
```
ChumakAn, line 11: another option is - creating a constructor in a Machine class with name and color and using the `super(name, color)` here

ChumakAn (approved):

PR #385, opened 2022-10-13
Implement Producers with PECS

Andew-Miroshnikov:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
...
+import core.mate.academy.model.Bulldozer;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
+    @Override
+    public List<Bulldozer> get() {
+        Bulldozer bulldozer1 = new Bulldozer();
+        Bulldozer bulldozer2 = new Bulldozer();
+        Bulldozer bulldozer3 = new Bulldozer();
+        List<Bulldozer> bulldozers = List.of(bulldozer1,bulldozer2,bulldozer3);
```
YuriiHryniv, line 12: You could also try to initialize those machines without the need of creating every single field

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+        if (type == Truck.class) {
+            machines = new TruckProducer();
+        } else if (type == Excavator.class) {
+            machines = new ExcavatorProducer();
+        } else if (type == Bulldozer.class) {
+            machines = new BulldozerProducer();
+        } else {
+            return new ArrayList<>();
+        }
+        return machines.get();
```
YuriiHryniv, line 27: Your code would never reach 27th line. I guess you should rethink your logic here

YuriiHryniv (changes requested):

YuriiHryniv (approved):

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,9 +5,15 @@
...
 public class Excavator extends Machine {
+    private int size;
+
     public Excavator() {
     }
 
+    public Excavator(int size) {
+        this.size = size;
+    }
+
```
ChumakAn, line 16: what about inheriting the parent's class fields? (Machine) and initializing it here?

ChumakAn: same for all child classes

src/main/java/core/mate/academy/service/BulldozerProducer.java
ChumakAn: +

ChumakAn: `
List<Bulldozer> bulldozers = List.of(new Bulldozer(10), new Bulldozer(20), new Bulldozer(30));
`

src/main/java/core/mate/academy/service/MachineServiceImpl.java
ChumakAn: +

ChumakAn (changes requested):
Good, let's make it better

PR #387, opened 2022-10-13
Solution

OleksiyMasaltsev:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+    private final BulldozerProducer bulldozerProducer = new BulldozerProducer();
+    private final ExcovatorProducer excovatorProducer = new ExcovatorProducer();
+    private final TruckProducer truckProducer = new TruckProducer();
```
muzykaVP, line 17: can we create these lists when we need to?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+        if (type == Bulldozer.class) {
+            machines = bulldozerProducer.get();
+        } else if (type == Excavator.class) {
+            machines = excovatorProducer.get();
+        } else if (type == Truck.class) {
+            machines = truckProducer.get();
+        } else {
+            machines = Collections.emptyList();
+        }
+        return new ArrayList<>(machines);
```
muzykaVP, line 32: can we return necessary list right away if we found it, without redundant checking?

muzykaVP (changes requested):
great, but i think we can make few improvement in your code

muzykaVP (approved):
Great!

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,11 +5,19 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    public static final int DEFAULT_WEIGHT = 30_000;
+    
+    private int weight = DEFAULT_WEIGHT;
+
```
ChumakAn, line 11: it's better to initiallize it in the constructor and don't hardcode any values

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,11 +5,19 @@
  * Do not remove no-args constructor
  */
 public class Excavator extends Machine {
+    public static final int DEFAULT_HEIGHT = 15;
+
+    private int height = DEFAULT_HEIGHT;
+
```
ChumakAn, line 11: same for values here

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -5,11 +5,19 @@
  * Do not remove no-field constructor
  */
 public class Truck extends Machine {
+    public static final int DEFAULT_SPEED = 60;
+
+    private int speed = DEFAULT_SPEED;
```
ChumakAn, line 10: same for values here

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -5,11 +5,19 @@
...
+
     public Truck() {
     }
 
+    public void setSpeed(int speed) {
+        this.speed = speed;
+    }
```
ChumakAn, line 17: you don't use setters anywhere, it's redundant

ChumakAn (changes requested):
Good! Let's make it better

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,11 +5,14 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private final int weight;
+
     public Bulldozer() {
+        weight = 30;
```
ChumakAn, line 11: you should leave the no-args constructor untouched, but create another one with the weight field and initialize it in the place you use it, not hardcoding the value here

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,11 +5,14 @@
  * Do not remove no-args constructor
  */
 public class Excavator extends Machine {
+    private final int height;
+
     public Excavator() {
+        height = 15;
```
ChumakAn, line 11: you should leave the no-args constructor untouched, but create another one with the height field and initialize it in the place you use it, not hardcoding the value here

ChumakAn (changes requested):

ChumakAn (approved):
I'll aprove this, but my suggestion was - create the constructor in the Machine class with two fields it has -> in the child class constructor call this super constructor, for example 
```
public Excavator(String name, String color, int size) {
        super(name, color);
        this.size = size;
    }
```
For all of the child classes

PR #391, opened 2022-10-16
solved jv pecs task

HryhorashPavlo:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+        for (int i = 0; i < machines.size(); i++) {
+            machines.set(i, value);
+        }
+    }
+
+    @Override
+    public void startWorking(List<? extends Machine> machines) {
+        for (Machine machine : machines) {
+            machine.doWork();
+        }
```
GHSpeedWagon, line 37: what will be do this mathod if machines list will be null

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
- * Your implementation of MachineService.
- */
-public class MachineServiceImpl {
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
```
GHSpeedWagon, line 9: maybe we can import it out of class

GHSpeedWagon (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+
+    @Override
+    public void startWorking(List<? extends Machine> machines) {
+        if (machines == null) {
+            throw new NullPointerException();
+        } else {
+            for (Machine machine : machines) {
+                machine.doWork();
+            }
+        }
```
GHSpeedWagon, line 41: i guess we cant do something if method != null

GHSpeedWagon (approved):

ChumakAn (approved):
LGTM

PR #394, opened 2022-10-17
Ivg ma

igorgalenko:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<? extends Machine> machines = new LinkedList<>();
+        if (type == Truck.class) {
+            machines = new TruckProducer().get();
```
YuriiHryniv, line 18: You could do that without the need of creating machine variable.

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.LinkedList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    @Override
+    public List<Excavator> get() {
+        List<Excavator> excavators = new LinkedList<>();
+        excavators.add(new Excavator());
+        excavators.add(new Excavator());
+        return excavators;
```
YuriiHryniv, line 13: It's also a great idea to return the value using static factory method

YuriiHryniv (changes requested):

YuriiHryniv (approved):
Seems good to me

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> { //useless T
```
ChumakAn, line 9: why is it useless? remove the comment

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Truck.class) {
+            return new LinkedList<>(new TruckProducer().get());
```
ChumakAn, line 17: what's the reason to use  Linked List?

ChumakAn (changes requested):
If you want to make your solution better, you should add some logic to the hierarchy of classes, also you can use super() in child classes to inherit the fields and make this class structure better-looking
image
Let's remove the redundant comment from the PR and make your code better👍

src/main/java/core/mate/academy/service/MachineService.java
igorgalenko: IDE marks this T like unused (grey color)

ChumakAn (approved):

PR #396, opened 2022-10-17
done all task from README.md

Danyl-Kniaziev:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,50 @@
...
+    private BulldozerProducer bulldozerProducer = new BulldozerProducer();
+    private ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+    private TruckProducer truckProducer = new TruckProducer();
+
+    @Override
+    public List<Machine> getAll(Class type) {
+        if (type.equals(Bulldozer.class)) {
+            List<Bulldozer> machines = bulldozerProducer.get();
+            machines.add(new Bulldozer());
+            return new ArrayList<>(machines);
```
igorgalenko, line 23: U can return list without creating machines variable

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,50 @@
...
+            List<Excavator> machines = excavatorProducer.get();
+            machines.add(new Excavator());
+            return new ArrayList<>(machines);
+        } else if (type.equals(Truck.class)) {
+            List<Truck> machines = truckProducer.get();
+            machines.add(new Truck());
+            return new ArrayList<>(machines);
+        }
+        List<? extends Machine> machines = truckProducer.get();
+        return new ArrayList<>(machines);
```
igorgalenko, line 34: Maybe it is not good to get list of trucks twice

igorgalenko (changes requested):

igorgalenko (approved):
LGTM

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,9 +5,33 @@
...
+    private int sizeBucket;
+    private int lengthHand;
+
     public Excavator() {
     }
 
+    public Excavator(int sizeBucket, int lengthHand) {
+        this.sizeBucket = sizeBucket;
+        this.lengthHand = lengthHand;
+    }
```
ChumakAn, line 17:
```suggestion
    public Excavator(String name, String color, int sizeBucket, int lengthHand) {
    	super(name, color);
        this.sizeBucket = sizeBucket;
        this.lengthHand = lengthHand;
    }
```

ChumakAn: what about inheriting the fields from super class (Machine)?

ChumakAn: same for all

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,10 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
```
ChumakAn, line 6:
```suggestion
public class BulldozerProducer implements MachineProducer<Bulldozer> {
```

ChumakAn: same for all

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 public interface MachineProducer {
```
ChumakAn, line 6:
```suggestion
public interface MachineProducer<T> {
```

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
...
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends Machine> get();
```
ChumakAn, line 11:
```suggestion
    List<T extends Machine> get();
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class type);
```
ChumakAn, line 20:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    private BulldozerProducer bulldozerProducer = new BulldozerProducer();
+    private ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+    private TruckProducer truckProducer = new TruckProducer();
```
ChumakAn, line 16: you can avoid creating this variables

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+
+    @Override
+    public List<Machine> getAll(Class type) {
+        if (type.equals(Bulldozer.class)) {
+            return new ArrayList<>(bulldozerProducer.get());
```
ChumakAn, line 21:
```suggestion
            return new ArrayList<>(new BulldozerProducer().get());
```

ChumakAn (changes requested):
Good! Let's make it better

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Danyl-Kniaziev: But we'll create new SomeProducer every time when we call method getAll()? Why is it better?

ChumakAn: because you're creating them all 3 in every case, wasting some memory 
for example, if you need Bulldozer list, it's no need to create other objects

ChumakAn (approved):
Approve, but remember that in case you had 10 Producer's you'll waste memory on each of them without no need

PR #398, opened 2022-10-18
Producers for Machines was created. Interfaces MachineService, Machin…

NSay0:
…eProducer was parametrized. Implementation of MachineService was added.

### DIFF
--- a/src/main/java/core/mate/academy/model/Bulldozer.java
+++ b/src/main/java/core/mate/academy/model/Bulldozer.java
@@ -1,11 +1,17 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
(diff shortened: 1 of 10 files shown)


yehbreus (approved):
LGFM 🧙🏿‍♂️

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,13 +1,19 @@
...
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
+    private int specialBulldozerField;
+
     public Bulldozer() {
     }
```
ChumakAn, line 8: it will be good decision to add one more constructor and use super method for the abstract class constructor(you can add it )

ChumakAn (approved):
Good, take a notice at the comment I have left

PR #400, opened 2022-10-18
create a MachineServiceImpl

Matolch:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
```
Bodddd, line 17: Your solution looks good but this code must be a good example of polymorphism.
Let's create an abstract variable of a type MachineProducer and initialize it with an applicable producer implementation. Then just call the method get one time.

Bodddd (changes requested):
Almost perfect!

Bodddd (approved):
Good job!

KostyaNikiforov (approved):

PR #404, opened 2022-10-19
hw-pecs

Iryna-a:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
-/**
- * Parameterize this service and add its implementation in a separate class.
- */
-public interface MachineService {
+public interface MachineService<T> {
```
AlinaOryshak, line 6: I think you should restrict types that can be used with this interface.

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
-     * Return the list of machines.
-     * In the implementation of this method please use your MachineProducer implementations
-     * See that 'Class type' is not parametrized.
-     * Consider to parametrize this Class< PARAMETRIZE_ME > with specific wildcard bounding
-     *
-     * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
-     * @return the list of machines
-     */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class type);
```
AlinaOryshak, line 8: Parametrize `Class< ...... > type` with specific wildcard bounding to pass in method only Machine sub classes

AlinaOryshak (changes requested):
Great !!! I have several suggestions

AlinaOryshak (approved):
LGTM ))

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,10 +1,9 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
+    private String bulldozerType;
+    private int weight;
+
```
ChumakAn, line 6: Optional, but it'll be a good solution to write a constructor in the Machine class and inherit it in the child classes like Bulldozer, Truck etc, using the keyword `super` in the new constructor

ChumakAn (changes requested):
Well done, minor comment

PR #424, opened 2022-11-17
Pecs solution

Kateryna-Takhtai:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +14,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
```
aaananas, line 17:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer<? extends Machine> machineProducer;
+        if (Bulldozer.class.equals(type)) {
+            machineProducer = new BulldozerProducer();
+        } else if (Truck.class.equals(type)) {
```
aaananas, line 17: lets create class level variables for each producer and invoke method get() here
to avoid creating a new instance each time this method is invoked

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+        } else {
+            return new ArrayList<>();
+        }
+        List<? extends Machine> machines = machineProducer.get();
+        return new ArrayList<>(machines);
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        Collections.fill(machines, value);
```
aaananas, line 30: lets use for loop and set value on i index

aaananas (changes requested):

aaananas (approved):

PR #427, opened 2022-11-19
jv-pecs solution

oksana-rudenko:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+            return new ArrayList<>();
+        }
+        MachineProducer<? extends Machine> machineProducer = new TruckProducer();
+        if (type == Bulldozer.class) {
+            machineProducer = new BulldozerProducer();
+        }
+        if (type == Excavator.class) {
+            machineProducer = new ExcavatorProducer();
+
+        }
```
aleksandr-bevza, line 25: lets create class-level fields with producers
don't create a new instance each time this method is invoked
```suggestion
        List<? extends Machine> machineList = new ArrayList<>();
        if (type == Bulldozer.class) {
            machineList = bulldozerProducer.get();
        }
        if (type == Excavator.class) {
            machineList = excavatorProducer.get();
        }
        if (type == Truck.class) {
            machineList = truckProducer,get();
        }
```

aleksandr-bevza (changes requested):

Rommelua (approved):

PR #440, opened 2022-12-03
pecs v.1

ItsUnbidden:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+        }
+        if (type == Excavator.class) {
+            producer = new ExcavatorProducer();
+            machines = producer.get();
+        }
+        if (type == Truck.class) {
+            producer = new TruckProducer();
+            machines = producer.get();
+        }
+        return new ArrayList<>(machines);
```
zachar-chorny, line 36: Why do you need the machines and producer variables when you can return them all at once in an if block?

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class TruckProducer implements MachineProducer<Truck> {
+    
+    @Override
+    public List<Truck> get() {
+        List<Truck> trucks = new ArrayList<>();
+        for (int i = 0; i < 3; i++) {
+            trucks.add(new Truck());
+        }
+        return trucks;
```
zachar-chorny, line 15: Return a list of objects with different data. Fix everywhere

zachar-chorny (changes requested):

zachar-chorny (approved):
Good job

PR #443, opened 2022-12-09
Implement MachineServiceImpl

Dmytro-Mk:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,36 @@
...
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<? extends Machine> machines =
+                type == Bulldozer.class ? new BulldozerProducer().get()
+                : type == Excavator.class ? new ExcavatorProducer().get()
+                : type == Truck.class ? new TruckProducer().get()
+                : Collections.emptyList();
+        return new ArrayList<>(machines);
```
KostyaNikiforov, line 22: This code must be a good example of polymorphism in java. Let's create a var of a type MachineProducer and call the method `get` only once.

KostyaNikiforov (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -14,11 +14,14 @@
...
     @Override
     public List<Machine> getAll(Class<? extends Machine> type) {
-        List<? extends Machine> machines =
-                type == Bulldozer.class ? new BulldozerProducer().get()
-                : type == Excavator.class ? new ExcavatorProducer().get()
-                : type == Truck.class ? new TruckProducer().get()
-                : Collections.emptyList();
+        if (type != Bulldozer.class && type != Excavator.class && type != Truck.class) {
+            return Collections.emptyList();
+        }
```
KostyaNikiforov, line 19: Actually, this is not a scalable code. Now, when there are more Cars in the app, you will need to change this expression also, and it will become longer and longer.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -14,11 +14,14 @@
...
+        MachineProducer<? extends Machine> machineProducer =
+                type == Bulldozer.class ? new BulldozerProducer()
+                : type == Excavator.class ? new ExcavatorProducer()
+                : new TruckProducer();
```
KostyaNikiforov, line 23: It is not the best practice to use a multilevel ternary operator. Usually, the switch-case construction is used to increase readability

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
🔥

PR #458, opened 2022-12-23
Jv pecs solution

v-fesenko:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,21 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private List<Bulldozer> machines = new ArrayList<>();
```
Rommelua, line 8: Create a new list every time
```suggestion
```

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
...
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<M extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends Machine> get();
```
Rommelua, line 11: Don't you know what exact type is inside? Do not use ? in returned objects - only in the arguments

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<M extends Machine> {
```
Rommelua, line 9: What is the reason for this parametrization if you do not use **M** type?

src/main/java/core/mate/academy/service/ProducerStrategy.java
```diff
@@ -0,0 +1,21 @@
...
+        if (type == Bulldozer.class) {
+            return new BulldozerProducer();
+        }
+        if (type == Excavator.class) {
+            return new ExcavatorProducer();
+        }
+        if (type == Truck.class) {
+            return new TruckProducer();
+        }
+        return null;
```
Rommelua, line 19: Do not return null - better throw an exception

src/main/java/core/mate/academy/service/ProducerStrategy.java
```diff
@@ -0,0 +1,21 @@
...
+
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+
+public class ProducerStrategy {
+    public MachineProducer<? extends Machine> getMachineProducer(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            return new BulldozerProducer();
```
Rommelua, line 11: Make Producers class-level variables.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,35 @@
...
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        ProducerStrategy strategy = new ProducerStrategy();
```
Rommelua, line 14: Make ProducerStrategy a class-level variable.

Rommelua (changes requested):

src/main/java/core/mate/academy/service/ProducerStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+        if (type == Bulldozer.class) {
+            return bulldozerProducer;
+        }
+        if (type == Excavator.class) {
+            return excavatorProducer;
+        }
+        if (type == Truck.class) {
+            return truckProducer;
+        }
+        throw new RuntimeException("Got invalid class for producer. Please, input valid class");
```
karma-o, line 23: I would use `Map<Class<? extends Machine>, MachineProducer<? extends Machine>>` instead of  `if`s, this way our code will be easy to scale

karma-o (changes requested):

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -8,6 +8,11 @@ public class Bulldozer extends Machine {
     public Bulldozer() {
     }
 
+    public Bulldozer(String name, String color) {
+        this.setName(name);
+        this.setColor(color);
```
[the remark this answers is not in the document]
dreeeamcatcher: That was my solution But the abstract class Machine has name and color as private fields so that I can access them only through inherited functions - getters & setters. 
Is it ok to change the code already provided within a template?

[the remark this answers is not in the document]
dreeeamcatcher: I deleted the custom constructor, as it is required by checkstyle to keep `name` and `color` fields as private ones. So I can't assign them directly in the subclass.

kozhukhovsky (approved):

PR #467, opened 2022-12-26
solved task

yaaariiik:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
-     * Return the list of machines.
-     * In the implementation of this method please use your MachineProducer implementations
-     * See that 'Class type' is not parametrized.
-     * Consider to parametrize this Class< PARAMETRIZE_ME > with specific wildcard bounding
-     *
-     * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
-     * @return the list of machines
-     */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
nacenik, line 8:
```suggestion
    List<Machine> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
-    /**
-     * Fill the machines list with passed value
-     * Replace the Object with parametrized value.
-     * This method should be able to work well with any type of machine passed as 'value'
-     *
-     * @param machines - list of machines to be filled with value
-     * @param value    - any object of machine sub class
-     */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, Machine value);
```
nacenik, line 10:
```suggestion
    void fill(List<? super T> machines, Machine value);
```

src/main/java/core/mate/academy/service/MachineService.java
nacenik: Fix it in all places in this interface

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,44 @@
...
+            return new ArrayList<>(machines);
+        }
+        List<Machine> list = new ArrayList<>();
+        return list;
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        int listSize = machines.size();
+        for (int i = 0; i < listSize; i++) {
```
nacenik, line 33:
```suggestion
        for (int i = 0; i < machines.size(); i++) {
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,44 @@
...
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            List<? extends Machine> machines = new BulldozerProducer().get();
```
nacenik, line 15: Create additional class variables for BulldozerProducer, ExcavatorProducer, and TruckProducer, then return simple bulldozerProducer.get(), ... instead of creating a new instance every time

nacenik (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
yaaariiik: I can't do this way, because my buldozerProducer.get() method will return List of Buldozers, truckProducer.get() will return List of Trucks etc. But this method should return List of Machines, it will not compile. That's why I create List with wildcard that extends any Machine class and return it as List of Machines, as it requires.

aleksandr-bevza (approved):

PR #471, opened 2023-01-01
Completed PECS 01-01-2023

ArtemMakovskyy:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -4,10 +4,30 @@
  * Add some custom fields that could be only in Bulldozer
  * Do not remove no-args constructor
  */
-public class Bulldozer extends Machine {
+
+public class Bulldozer extends Machine implements Workable {
```
Rommelua, line 8: why implements Workable ???
```suggestion
public class Bulldozer extends Machine {
```

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,28 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
+    private List<Bulldozer> list;
+
+    {
```
Rommelua, line 10: never use init blocks - it's a bad practice.
In this task better create a new list in the method.

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,13 @@
...
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
+     *
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends Machine> get();
```
Rommelua, line 12: Never use wild cards when you return something. Wild cards for income parameters validation

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
```
Rommelua, line 9: Think about: why do you need this **T** at all?

Rommelua (changes requested):

PR #475, opened 2023-01-04
hw-2-sulution

nick0226:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
```
aaananas, line 17:
```suggestion
        if (type == Bulldozer.class) {
            return new ArrayList<>(bulldozerProducer.get());
```
better create class variables to avoid creating a new producer each time this method is invoked
remember to use abstract reference type

aaananas (approved):

PR #476, opened 2023-01-04
Added additional fields for classes Truck, Excavator, Bulldozer.

nkharyshyn:
Parameterized MachineProducer and MachineService with suitable option to strict type. Created needed lists in implementations of MachineProducer using constructors of certain classes. Added necessary logic to methods of MachineServiceImpl class

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineServiceImpl.java
+++ b/src/main/java/core/mate/academy/service/MachineServiceImpl.java
@@ -1,7 +1,42 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 /**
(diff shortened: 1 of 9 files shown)


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            List<Bulldozer> bulldozers = new BulldozerProducer().get();
+            return new ArrayList<>(bulldozers);
```
aaananas, line 19:
```suggestion
        if (type == Bulldozer.class) {
            List<Bulldozer> bulldozers = bulldozerProducer.get();
            return new ArrayList<>(bulldozers);
```
you can make producers a class level variable to avoid creating a new instance each time this method is invoked

aaananas (approved):
great job

PR #477, opened 2023-01-07
implemented specific interface with wild card

Oleksii-Bilko:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (Bulldozer.class.equals(type)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        }
+        if (Excavator.class.equals(type)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        if (Truck.class.equals(type)) {
+            return new ArrayList<>(new TruckProducer().get());
+        }
```
karma-o, line 22: can we not create a new Producers each time we are calling `getAll()` method? Let's move them to a class level variables or maybe even a map of types like this
`Map<Class<? extends Machine>, MachineProducer<? extends Machine>> producers`
to avoid `if`s

karma-o (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Oleksii-Bilko: Map is the next chapter :)

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
+    void fill(List<? super T> machines, T value);
 
-    /**
-     * Call the method doWork() from each machine.
-     * This method should be able to accept a list of bulldozers as well as list of trucks.
-     *
-     * @param machines - the list of machines
-     */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
aaananas, line 12:
```suggestion
    void startWorking(List<? extends T> machines);
```

aaananas (approved):

PR #482, opened 2023-01-15
implemented MachineProducer implementations and implmentation of Mach…

YevheniiKilovyi:
…ineService


aleksandr-bevza (approved):
Good job!

PR #486, opened 2023-01-19
Make implementations MachineProducer and MachineService metods

Yulia-Kas:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<? extends Machine> machines;
+        MachineProducer<? extends Machine> machineProducer;
+        if (type.equals(Bulldozer.class)) {
+            machineProducer = new BulldozerProducer();
```
aaananas, line 19:
```suggestion
        if (type.equals(Bulldozer.class)) {
            return new ArrayList<>(bulldozerProducer.get(machines));
```
lets create 3 class variables - to avoid creating new instances each time this method is invoked

aaananas (approved):

PR #495, opened 2023-01-27
Make necessary classes and implementations

katerynapoli:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,9 +5,43 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private String trackType;
+    private String cabType;
+    private int operatingWeight;
+
     public Bulldozer() {
     }
```
okuzan, line 13: delete unused constructors everywhere

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        bulldozers.add(new Bulldozer("Track type 1", "Cab type 1", 100));
+        bulldozers.add(new Bulldozer("Track type 2", "Cab type 2", 200));
+        bulldozers.add(new Bulldozer("Track type 3", "Cab type 3", 300));
+        return bulldozers;
```
okuzan, line 14: we can do it with one liner using List.of()

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        }
+        if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        if (type == Truck.class) {
+            return new ArrayList<>(new TruckProducer().get());
+        }
+        return new ArrayList<>();
```
okuzan, line 25: use hashmap with class as key to do this in more elegant way

okuzan (changes requested):
Nice code!

src/main/java/core/mate/academy/model/Bulldozer.java
katerynapoli: As I understand, I'm not allowed to do this, because no-arg constructors are used in the tests

src/main/java/core/mate/academy/service/MachineServiceImpl.java
katerynapoli: I'm not familiar with hashmap yet, it will be the next topic, as I see

andriy-shulepa (approved):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
okuzan: You're right, sorry

PR #503, opened 2023-02-07
PECS solution

serhii-shyian:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,47 @@
...
+        }
+        if (type == Bulldozer.class) {
+            BulldozerProducer dozerProducer = new BulldozerProducer();
+            machines = new ArrayList<>(dozerProducer.produce());
+        }
+        if (type == Truck.class) {
+            TruckProducer truckProducer = new TruckProducer();
+            machines = new ArrayList<>(truckProducer.produce());
+        }
+        return machines;
```
aaananas, line 31: lets create class level variables for producers - to avoid creating a new instance each time this method is invoked
remember about OOP principles - use abstract reference type

aaananas (approved):

PR #507, opened 2023-02-09
feat: jv-pecs #comment pecs implementation

sania517:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,32 @@
...
+        secondBulldozer.setColor("blue");
+        thirdBulldozer.setColor("green");
+        list.add(firstBulldozer);
+        list.add(secondBulldozer);
+        list.add(thirdBulldozer);
+    }
+
+    @Override
+    public List<Bulldozer> get() {
+        return list;
```
okuzan, line 30: this can be done with less code. Instead of having constructor and field, just return directly: `return List.of(new Bulldozer("first", "red", ...))` Also you haven't committed changes to Bulldozer, Excavator, Truck, create constructors if you don't have them. Fix everywhere

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
```
okuzan, line 5:
```suggestion
public interface MachineProducer<T extends Machine> {
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+import java.util.Map;
+
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+    private static Map<Class<?>, MachineProducer> producerMap = new HashMap();
+
+    static {
+        producerMap.put(Bulldozer.class, new BulldozerProducer());
+        producerMap.put(Excavator.class, new ExcavatorProducer());
+        producerMap.put(Truck.class, new TruckProducer());
+    }
```
okuzan, line 19: this is okay, just FYI, in the real world this would likely be done with Map.of()

okuzan (changes requested):

aleksandr-bevza (approved):

PR #508, opened 2023-02-11
Solved PECS task

WallRun:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
aaananas, line 20:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer<? extends Machine> producer;
+        if (type == Bulldozer.class) {
+            producer = new BulldozerProducer();
+        } else if (type == Excavator.class) {
+            producer = new ExcavatorProducer();
+        } else if (type == Truck.class) {
+            producer = new TruckProducer();
+        } else {
+            return new LinkedList<>();
```
aaananas, line 22: you can make machinge producers  class level fields, this way you would avoid creation on a new instance each time this method is invoked

aaananas (changes requested):

aaananas (approved):

PR #522, opened 2023-02-21
Hw-pecs

KosMarch:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private Random random = new Random();
+
+    @Override
+    public List<? extends Machine> get() {
+        Bulldozer bulldozer1 = new Bulldozer(random.nextInt());
+        Bulldozer bulldozer2 = new Bulldozer(random.nextInt());
+        Bulldozer bulldozer3 = new Bulldozer(random.nextInt());
+        List<Bulldozer> bulldozers = new ArrayList<>();
```
denys-mg, line 17: probably it's better to move list of machines as a class variable, to not create it in get method

denys-mg (approved):
Good work!)

src/main/java/core/mate/academy/service/BulldozerProducer.java
nacenik: That's a bad idea, cause the class should contain fields that describe it. For example, can bulldozer contain a list of bulldozers inside in real life? Probably not.

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,23 @@
...
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private Random random = new Random();
+
+    @Override
+    public List<? extends Machine> get() {
```
nacenik, line 13: In producer realization, you should have the exact type of list, not the wildcard. Fix it in all producers

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,9 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
```
nacenik, line 9: don't delete Javadoc. it's not a simple comment. Return them everywhere

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
-    /**
-     * Fill the machines list with passed value
-     * Replace the Object with parametrized value.
-     * This method should be able to work well with any type of machine passed as 'value'
-     *
-     * @param machines - list of machines to be filled with value
-     * @param value    - any object of machine sub class
-     */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, core.mate.academy.model.Machine value);
```
nacenik, line 10: the second parameter should be generic type too.

nacenik (changes requested):

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public List<Bulldozer> get() {
+        Bulldozer bulldozer1 = new Bulldozer(random.nextInt());
+        Bulldozer bulldozer2 = new Bulldozer(random.nextInt());
+        Bulldozer bulldozer3 = new Bulldozer(random.nextInt());
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        bulldozers.add(bulldozer1);
+        bulldozers.add(bulldozer2);
+        bulldozers.add(bulldozer3);
+        return bulldozers;
```
karma-o, line 20: not a mistake but in the future `List.of()` method would be perfect here

karma-o (approved):
GJ, left some suggestion

PR #529, opened 2023-02-26
Implemented task using PECS 

copypasterr:


VitaliiPotseluiko (approved):

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,7 @@
 
 import java.util.List;
 
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<T> {
```
karma-o, line 5: Should it be able to accept types of `String` or `Integer`? Because your implementation has it

karma-o (changes requested):

sarakhmen (approved):
Good job :)

PR #533, opened 2023-03-01
HomeWork JavaCore/PECS

NickolayVoloshchuk:


Deschna (approved):
You used some constructions that i didn't understand properly, but all else good, have nothing to change :P

nacenik (approved):
Great job! You can improve your solution if you want 😉

PR #537, opened 2023-03-05
PECS practice

Deschna:


NickolayVoloshchuk (approved):
Good job! It's interesting that sometimes you could see how to solve the same problem in a different way ;)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<? extends Machine> machines;
+        if (Bulldozer.class.equals(type)) {
+            machines = new BulldozerProducer().get();
+        } else if (Excavator.class.equals(type)) {
+            machines = new ExcavatorProducer().get();
+        } else if (Truck.class.equals(type)) {
+            machines = new TruckProducer().get();
```
petro-ushchuk, line 20: No need to create these instances every time you call `getAll` method

petro-ushchuk (approved):
Good work! Read comment bellow 🔥

PR #539, opened 2023-03-07
solved task

slaybrute:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,13 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<M extends Machine> {
```
uliana-andrushchuk, line 6: The common practice in Java is to use `T` for generics

uliana-andrushchuk (approved):
Added minor comments

PR #545, opened 2023-03-13
Hw jv pecs

vromanyuk:
added methods set and get for better structure (is can take parameters of fields or added) but they don't use in this structure

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineProducer.java
+++ b/src/main/java/core/mate/academy/service/MachineProducer.java
@@ -1,13 +1,11 @@
 package core.mate.academy.service;
 
-import core.mate.academy.model.Machine;
-
 import java.util.List;
 
-public interface MachineProducer<Machine> {
+public interface MachineProducer<E> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<? extends Machine> get();
+    List<? extends E> get();
 }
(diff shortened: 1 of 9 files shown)


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<E> {
```
aaananas, line 5:
```suggestion
public interface MachineProducer<T extends Machine> {
```

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
...
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends E> get();
```
aaananas, line 10:
```suggestion
    List<T> get();
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -5,7 +5,7 @@
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<E> {
```
aaananas, line 8: lets add upper bound

aaananas (changes requested):

okuzan (approved):
Nice work

PR #550, opened 2023-03-19
jv-pecs-solution

knyazkyylubomir:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<? extends Machine> machines = new ArrayList<>();
+        if (type == Bulldozer.class) {
+            machines = new BulldozerProducer().get();
+        }
+        if (type == Truck.class) {
+            machines = new TruckProducer().get();
+        }
+        if (type == Excavator.class) {
+            machines = new ExcavatorProducer().get();
```
olekskov, line 24: services class rely on should be private final variables on class level not to be recreated each time we call some method

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        for (int i = 0; i < machines.size(); i++) {
+            machines.set(i, value);
+        }
+    }
+
+    @Override
+    public void startWorking(List<? extends Machine> list) {
+        for (int i = 0; i < list.size(); i++) {
+            list.get(i).doWork();
+        }
```
olekskov, line 40: for-each cycle will look more tidy and readable

olekskov (approved):
Good job with theme, approved. Please pay attention to comments below for further tasks

PR #556, opened 2023-04-01
added changes

oooRule:


Ivan95kos (changes requested):
Task is not implemented

olekskov (changes requested):
Build is failed, please run 'mvn clean package' command locally and fix problems highlighted in report

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
```
uliana-andrushchuk, line 20:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines,T value);
```
uliana-andrushchuk, line 30:
```suggestion
    void fill(List<? super T> machines, T value);
```

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,6 +5,16 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private String bullColor;
+    private int bullHeight;
+    private int bullLength;
```
uliana-andrushchuk, line 10: It's not a good practice to use such names for class fields (better to name it without `bull` prefix). As you already have `color` field in class Machine you can replace it with more type-specific field

uliana-andrushchuk: The same for other classes

uliana-andrushchuk (changes requested):

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,6 +5,16 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private String name;
+    private int height;
+    private int length;
+
+    public Bulldozer(String bullColor, int bullHeight, int bullLength) {
```
uliana-andrushchuk, line 12:
```suggestion
    public Bulldozer(String name, int height, int length) {
```

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,6 +5,16 @@
  * Do not remove no-args constructor
  */
 public class Excavator extends Machine {
+    private String name;
+    private int height;
+    private int length;
+
```
uliana-andrushchuk, line 11: It would be better to use some specific field for each class, not the same

uliana-andrushchuk (changes requested):

kozhukhovsky (approved):

PR #557, opened 2023-04-01
executed

VolodymyrDavidov:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -8,6 +8,10 @@ public class Bulldozer extends Machine {
     public Bulldozer() {
     }
 
+    public Bulldozer(String name, String color, int horsePower) {
+        super();
```
uliana-andrushchuk, line 12: Values of fields `name`, `color` and `horsePower` are just ignored.
Please, create  constructor with 3 parameters in `Machine` class and set it (ex this. name = name; ...)

uliana-andrushchuk (changes requested):

olekskov (approved):

PR #565, opened 2023-04-10
solve task

z1row:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
...
+    void fill(List<? super Machine> machines, Machine value);
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
okuzan, line 38: in methods of this class, use T instead of Machine everywhere

okuzan (changes requested):
Good job!

okuzan (approved):

PR #571, opened 2023-04-15
all tests passed

sshestaka:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,50 @@
...
+        if (type.equals(Excavator.class)) {
+            listMachines = excavatorProducer.get();
+            return new ArrayList<>(listMachines);
+        } else if (type.equals(Bulldozer.class)) {
+            listMachines = bulldozerProducer.get();
+            return new ArrayList<>(listMachines);
+        } else if (type.equals(Truck.class)) {
+            listMachines = truckProducer.get();
+            return new ArrayList<>(listMachines);
+        }
```
Ivan95kos, line 37: You can also rewrite like this:
```suggestion
        if (type.equals(Excavator.class)) {
            return new ArrayList<>(excavatorProducer.get());
        } else if (type.equals(Bulldozer.class)) {
            return new ArrayList<>(bulldozerProducer.get());
        } else if (type.equals(Truck.class)) {
            return new ArrayList<>(truckProducer.get());
        }
```

Ivan95kos (approved):

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Truck;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+
```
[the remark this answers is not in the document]
sshestaka: Ok!

src/main/java/core/mate/academy/service/MachineServiceImpl.java
sshestaka: Yes, thank you!

PR #583, opened 2023-05-03
task solved

MykytaKat:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+                return new ArrayList<>(machineProducer.get());
+            case "Truck":
+                machineProducer = new TruckProducer();
+                return new ArrayList<>(machineProducer.get());
+            case "Excavator":
+                machineProducer = new ExcavatorProducer();
+                return new ArrayList<>(machineProducer.get());
+            default:
+                return new ArrayList<>();
+        }
```
sarakhmen, line 23: I'd recommend creating a map as a class field (using class as a key and producer as a value) and using this map instead of switch

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+                machineProducer = new BulldozerProducer();
+                return new ArrayList<>(machineProducer.get());
+            case "Truck":
+                machineProducer = new TruckProducer();
+                return new ArrayList<>(machineProducer.get());
+            case "Excavator":
+                machineProducer = new ExcavatorProducer();
+                return new ArrayList<>(machineProducer.get());
+            default:
+                return new ArrayList<>();
```
sarakhmen, line 22:
```suggestion
                return Collections.emptyList();
```

sarakhmen (changes requested):

rovein (approved):
Good job!

PR #585, opened 2023-05-03
I created different classes of representation of machines, as well as…

V-Solianyk:
… their representations regarding their creation and addition to the list, implemented a class in which any type will come as a parameter, and the necessary implementation will be dynamically created.

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineServiceImpl.java
+++ b/src/main/java/core/mate/academy/service/MachineServiceImpl.java
@@ -1,7 +1,35 @@
 package core.mate.academy.service;
 
-/**
- * Your implementation of MachineService.
- */
-public class MachineServiceImpl {
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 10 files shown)


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,35 @@
...
+        switch (type.getSimpleName()) {
+            case "Truck":
+                return new ArrayList<>(new TruckProduce().get());
+            case "Excavator":
+                return new ArrayList<>(new ExcavatorProduce().get());
+            case "Bulldozer":
+                return new ArrayList<>(new BulldozerProducer().get());
+            default:
+                return new ArrayList<>();
+        }
```
sarakhmen, line 19: I'd like to leave here several points: 
- I'd recommend creating a map as a class field (using class as a key and producer as a value) and using this map instead of switch

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #589, opened 2023-05-07
implemented MachineService and MachineProducers

nechupara:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Machine> {
+    public List<Machine> get() {
+        List<Machine> bulldozers = new ArrayList<>();
+        bulldozers.add(new Bulldozer());
+        bulldozers.add(new Bulldozer());
+        bulldozers.add(new Bulldozer());
+        return bulldozers;
```
sarakhmen, line 14: I'd recommend simply using List.of() here

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,18 @@
...
+public class ExcavatorProducer implements MachineProducer<Machine> {
+    public List<Machine> get() {
+        List<Machine> excavators = new ArrayList<>();
+        excavators.add(new Excavator());
+        excavators.add(new Excavator());
+        excavators.add(new Excavator());
+        return excavators;
+    }
+
+    ;
```
sarakhmen, line 17:
```suggestion
```

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
```
sarakhmen, line 9: Please remove all redundant comments

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -5,7 +5,7 @@
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
sarakhmen, line 8:
```suggestion
public interface MachineService<T extends Machine> {
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +16,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class type);
```
sarakhmen, line 19:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        }
+        if (type == Excavator.class) {
+            return new ExcavatorProducer().get();
+        }
+        if (type == Truck.class) {
+            return new TruckProducer().get();
+        } else {
+            return new ArrayList<>();
+        }
+
```
sarakhmen, line 26: I'd like to leave here several points:

sarakhmen (changes requested):
Good job, but let's improve your solution :)

PR #590, opened 2023-05-08
add all needed classes and methods body, and parameterized generics

ykhmaruk:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
-public class MachineServiceImpl {
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+
```
sarakhmen, line 11:
```suggestion
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else if (type == Truck.class) {
+            return new ArrayList<>(new TruckProducer().get());
+        } else {
+            return new ArrayList<>();
+        }
```
sarakhmen, line 22: I'd like to leave here several points:

sarakhmen (changes requested):

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #592, opened 2023-05-08
Solved task, with using PECS

Pkrasota:


src/main/java/core/mate/academy/service/BuildozerProducer.java
```diff
@@ -0,0 +1,19 @@
...
+public class BuildozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        Bulldozer firstBulldozer = new Bulldozer();
+        Bulldozer secondBulldozer = new Bulldozer();
+        firstBulldozer.setName("John Deere");
+        firstBulldozer.setColor("Yellow");
+        secondBulldozer.setName("CAT");
+        secondBulldozer.setColor("Orange");
+        return new ArrayList<>(List.of(firstBulldozer, secondBulldozer));
```
sarakhmen, line 16: Please pass parameters to the constructor and consider simply using List.of() here, so that this method is much easier to read. Fix the same issue in all places

src/main/java/core/mate/academy/service/BuildozerProducer.java
```diff
@@ -0,0 +1,19 @@
...
+    }
+}
+
```
sarakhmen, line 19:
```suggestion
```

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,19 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+
```
sarakhmen, line 8:
```suggestion
```

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
...
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
```
sarakhmen, line 10: Please fix all redundant comments in your solution ;)

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
sarakhmen, line 20:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, Machine value);
```
sarakhmen, line 30:
```suggestion
    void fill(List<? super T> machines, T value);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
...
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
sarakhmen, line 38:
```suggestion
    void startWorking(List<? extends T> machines);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+                machineProducer = new BuildozerProducer();
+                return new ArrayList<>(machineProducer.get());
+            case "Excavator":
+                machineProducer = new ExcavatorProducer();
+                return new ArrayList<>(machineProducer.get());
+            case "Truck":
+                machineProducer = new TruckProducer();
+                return new ArrayList<>(machineProducer.get());
+            default:
+                return new ArrayList<>();
```
sarakhmen, line 25: I'd like to leave here several points:
- I'd recommend creating a map as a class field (using class as a key and producer as a value) and using this map instead of switch

sarakhmen (changes requested):

src/main/java/core/mate/academy/service/BuildozerProducer.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        for (int i = 0; i < bulldozerNames.length; i++) {
+            Bulldozer bulldozer = new Bulldozer();
+            bulldozer.setName(bulldozerNames[i]);
+            bulldozer.setColor(bulldozerColors[i]);
+            bulldozers.add(bulldozer);
+        }
+        return bulldozers;
```
sarakhmen, line 20: Actually, this is slightly overcomplicated. Consider doing following:
```suggestion
        List<Bulldozer> bulldozers = new ArrayList<>();
        for (int i = 0; i < bulldozerNames.length; i++) {
            Bulldozer bulldozer = new Bulldozer();
            bulldozer.setName(bulldozerNames[i]);
            bulldozer.setColor(bulldozerColors[i]);
            bulldozers.add(bulldozer);
        }
        return List.of(new Bulldozer("John Deere", "Yellow"), ...);
```
please fix it in all places

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
-/**
- * Parameterize this service and add its implementation in a separate class.
- */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
```
sarakhmen, line 7:
```suggestion
```

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :)

src/main/java/core/mate/academy/service/BuildozerProducer.java
Pkrasota: I'm not sure if I did exactly that

src/main/java/core/mate/academy/service/BuildozerProducer.java
```diff
@@ -0,0 +1,22 @@
...
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        for (int i = 0; i < bulldozerNames.length; i++) {
+            Bulldozer bulldozer = new Bulldozer();
+            bulldozer.setName(bulldozerNames[i]);
+            bulldozer.setColor(bulldozerColors[i]);
+            bulldozers.add(bulldozer);
+        }
+        return List.of(bulldozers.toArray(new Bulldozer[0]));
+    }
+}
```
sarakhmen, line 22: Actually my previous comment was confusing :( Idk whether I used 'suggestion' wrongly or this function just worked incorrectly. So, let's repeat one more time :)
```suggestion
public class BuildozerProducer implements MachineProducer<Bulldozer> {
    @Override
    public List<Bulldozer> get() {
        return List.of(new Bulldozer(...), new Bulldozer(...));
    }
}
```
But I just noticed that you don't have an overridden constructor. In this case, your first solution is also applicable :)

sarakhmen (approved):
Good job :)

PR #593, opened 2023-05-09
Created an implementation of MachineService and 3 "Producer" classes

viprko:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,27 @@
...
+        bulldozerD4 = new Bulldozer("VPAT", 3.81, 14008);
+        bulldozerD4.setName("Cat D4");
+        bulldozerD4.setColor("Yellow");
+        bulldozerD5 = new Bulldozer("Semi-Universal", 4.3, 17340);
+        bulldozerD5.setName("Cat D5");
+        bulldozerD5.setColor("Yellow");
+        bulldozerD6Lgp = new Bulldozer("Semi-Universal", 5.8, 22870);
+        bulldozerD6Lgp.setName("Cat D6");
+        bulldozerD6Lgp.setColor("Yellow");
+    }
```
sarakhmen, line 21:
```suggestion
```
Let's not create separate fields for your instances. I'd recommend directly creating them in get() method. Also you can pass parameters directly to the constructor so that setters are not used. Please fix the same issue in all places

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
+            TruckProducer truckProducer = new TruckProducer();
+            return new ArrayList<>(truckProducer.get());
+        } else if (type.equals(Bulldozer.class)) {
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
+            return new ArrayList<>(bulldozerProducer.get());
+        } else if (type.equals(Excavator.class)) {
+            ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+            return new ArrayList<>(excavatorProducer.get());
+        }
+        return new ArrayList<>();
```
sarakhmen, line 24: I'd like to leave here several points:

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        Collections.fill(machines, value);
```
sarakhmen, line 29: let's implement your own fill method :))

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :)

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :)

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
...
+import core.mate.academy.model.Bulldozer;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        Bulldozer bulldozerD4 = new Bulldozer("Cat D4", "Yellow", "VPAT", 3.81, 14008);
+        Bulldozer bulldozerD5 = new Bulldozer("Cat D5", "Yellow","Semi-Universal", 4.3, 17340);
+        Bulldozer bulldozerD6Lgp = new Bulldozer("Cat D6LGP", "Yellow", "Semi-Universal",
+                5.8, 22870);
```
sarakhmen, line 12: actually here we can avoid creating these variables and directly create buldozers inside List.of()

sarakhmen (approved):
Good job ;)

PR #602, opened 2023-05-18
jv-pecs

s-hrynus:
…task.

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineService.java
+++ b/src/main/java/core/mate/academy/service/MachineService.java
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
 
     /**
      * Return the list of machines.
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
 
     /**
      * Fill the machines list with passed value
@@ -26,7 +27,7 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, Machine value);
 
     /**
      * Call the method doWork() from each machine.
@@ -34,5 +35,5 @@ public interface MachineService {
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
 }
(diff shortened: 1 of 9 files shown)


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
okuzan, line 20:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, Machine value);
```
okuzan, line 30:
```suggestion
    void fill(List<? super T> machines, T value);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
...
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
okuzan, line 38:
```suggestion
    void startWorking(List<? extends T> machines);
```

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,14 @@
...
+import core.mate.academy.model.Truck;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+    @Override
+    public List<Truck> get() {
+        Truck truck1 = new Truck("tanker", 112233, 20);
+        Truck truck2 = new Truck("tanker", 445566, 25);
+        Truck truck3 = new Truck("tanker", 778899, 50);
+        return List.of(truck1, truck2, truck3);
```
okuzan, line 12:
```suggestion
        return List.of(new Truck("tanker", 778899, 50),
```
You can simplify the code by directly creating new instances of the Truck class within the List.of method call, without the need for storing them in temporary variables

s-hrynus: in this task totally agree. But in future, if I use these variables somewhere, it'll better to store them?

olekskov (approved):
Good job

src/main/java/core/mate/academy/service/TruckProducer.java
olekskov: You may introduce variables if they are reused or if they add clarity to code. So for sure there are possible situations where creation of variable will provide benefits.
But if not then it's better to use less variables :)

PR #607, opened 2023-05-26
Hometask changes completed 

MarynaKushniruk:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,11 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
```
Ivan95kos, line 6: It is better to set type of Producer
```suggestion
public class BulldozerProducer implements MachineProducer<Bulldozer> {
```

Ivan95kos (approved):

PR #610, opened 2023-06-07
done pecs

aTiukhtii:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,11 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        return List.of(new Bulldozer(), new Bulldozer(), new Bulldozer());
```
overpathz, line 9: Nice, but what if we want to add 100 objects?
I don't know whether it's required in this hw, but I'll leave this minor remark.
In any way, instead of doing one thing many times (you are creating the same objects 3 times), you can use loops.

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,11 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Truck;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer {
```
overpathz, line 6: MachineProducer is parameterized class.
Raw type here.

overpathz (changes requested):
Great. Provide fix to minor remarks and it will be good to approve

src/main/java/core/mate/academy/service/BulldozerProducer.java
aTiukhtii: for this hw is always 3, so the loop is not necessary

PR #613, opened 2023-06-14
solution

OlehSarapuk:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+    private final BulldozerProducer bulldozerProducer = new BulldozerProducer();
+    private final ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+    private final TruckProducer truckProducer = new TruckProducer();
+
+    @Override
+    public List<T> getAll(Class<? extends Machine> type) {
```
yevheniisynytsia, line 16:
```suggestion
    public List<T> getAll(Class<? extends T> type) {
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+        if (type.equals(Bulldozer.class)) {
+            return (List<T>) bulldozerProducer.get();
```
yevheniisynytsia, line 18:
```suggestion
            return (List<? extends T>) bulldozerProducer.get();
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        for (int i = 0; i < machines.size(); i++) {
+            machines.set(i, value);
+        }
+    }
+
+    @Override
+    public void startWorking(List<? extends Machine> machines) {
```
yevheniisynytsia, line 37:
```suggestion
    public void startWorking(List<? extends T> machines) {
```

yevheniisynytsia (changes requested):
In general, good job, but you should reuse the type you specify on the class level

PR #616, opened 2023-06-16
Added hw solution

KyrylH:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        switch (type.getSimpleName()) {
+            case "Bulldozer":
+                return new ArrayList<>(bulldozerProducer.get());
+            case "Excavator":
+                return new ArrayList<>(excavatorProducer.get());
+            case "Truck":
+                return new ArrayList<>(truckProducer.get());
```
karma-o, line 26: I'd replace it with the strategy `Map<Class<? extends Machine>, MachineProducer<? extends Machine>>` instead of switch based on class name string

karma-o (changes requested):
Looks good, just a couple of advanced comments since we already know what strategy is and other stuff :)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,37 +1,35 @@
...
+    private final Map<Class<? extends Machine>, MachineProducer<? extends Machine>>
+            producers = new HashMap<>();
 
     public MachineServiceImpl() {
-        bulldozerProducer = new BulldozerProducer();
-        excavatorProducer = new ExcavatorProducer();
-        truckProducer = new TruckProducer();
+        producers.put(Bulldozer.class, new BulldozerProducer());
+        producers.put(Excavator.class, new ExcavatorProducer());
+        producers.put(Truck.class, new TruckProducer());
```
karma-o, line 20: we should not have hardcode or any logic in the constructor, initialize it in the field with `Map.of()` as a constant, or inject the map in the constructor from outside of this method

karma-o (approved):
Approved with a comment

PR #628, opened 2023-06-30
Implementation of MachineProducer and MachineService interfaces

Olga-Melnichenko:
Implemented the MachineProducer interface with three implementations: BulldozerProducer, ExcavatorProducer, and TruckProducer. Each implementation returns a list of specific machines. Parameterized the MachineProducer interface, replacing Object in the get() method signature with the appropriate parameter. Created the MachineService interface and also parameterized it, replacing Object in the method signatures with the appropriate parameter using PECS. Implemented the MachineServiceImpl class, which implements the MachineService interface. Implemented the methods getAll(Class<? extends Machine> type), fill(List<? super Machine> machines, Machine value), and startWorking() according to the task requirements. The getAll(Class type) method selects the corresponding MachineProducer implementation based on the input class and calls its get() method to obtain the corresponding machines. The fill method updates all elements in the machines list with the passed value, which can be any subtype of Machine. The startWorking() method calls the doWork() method for each machine in the list. Ensured proper parameterization of interfaces and classes to restrict the types that can be used with them.

### DIFF
--- a/src/main/java/core/mate/academy/model/Bulldozer.java
+++ b/src/main/java/core/mate/academy/model/Bulldozer.java
@@ -1,11 +1,41 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
+    private String brand;
+    private String fuelType;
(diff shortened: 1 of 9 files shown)


okuzan (approved):
well done

PR #632, opened 2023-07-03
Initial implementation

barbequeue:
This is a complete nightmare! Generics, wildcards and PECS are something i would never understand! Java is something i would never understand!


olekskov (approved):
Good job

PR #648, opened 2023-07-13
created producers, set parametrization for methods, created implement…

Someboty:
…ation for service methods

### DIFF
--- a/src/main/java/core/mate/academy/model/Bulldozer.java
+++ b/src/main/java/core/mate/academy/model/Bulldozer.java
@@ -1,11 +1,13 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
+    private String bladeType;
+
     public Bulldozer() {
+    }
+
+    public void setBladeType(String bladeType) {
+        this.bladeType = bladeType;
     }
 
     @Override
--- a/src/main/java/core/mate/academy/service/BulldozerProducer.java
+++ b/src/main/java/core/mate/academy/service/BulldozerProducer.java
@@ -1,0 +1,32 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer<E> implements MachineProducer<E> {
+    @Override
+    public List<? extends Machine> get() {
+        Bulldozer firstBulldozer = new Bulldozer();
+        firstBulldozer.setColor("red");
+        firstBulldozer.setName("Caterpillar");
+        firstBulldozer.setBladeType("S blade");
+
+        Bulldozer secondBulldozer = new Bulldozer();
+        secondBulldozer.setColor("yellow");
+        secondBulldozer.setName("Komatsu");
+        secondBulldozer.setBladeType("U blade");
+
+        Bulldozer thirdBulldozer = new Bulldozer();
+        thirdBulldozer.setColor("black");
+        thirdBulldozer.setName("Shantui");
+        thirdBulldozer.setBladeType("S-U blade");
+
+        ArrayList<Bulldozer> bulldozers = new ArrayList<>();
+        bulldozers.add(firstBulldozer);
+        bulldozers.add(secondBulldozer);
+        bulldozers.add(thirdBulldozer);
+        return bulldozers;
+    }
+}
--- a/src/main/java/core/mate/academy/service/ExcavatorProducer.java
+++ b/src/main/java/core/mate/academy/service/ExcavatorProducer.java
@@ -1,0 +1,32 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer<E> implements MachineProducer<E> {
+    @Override
+    public List<? extends Machine> get() {
+        Excavator firstExcavator = new Excavator();
+        firstExcavator.setColor("red");
+        firstExcavator.setName("JCB");
+        firstExcavator.setBucketVolume(2.49);
+
+        Excavator secondExcavator = new Excavator();
+        secondExcavator.setColor("yellow");
+        secondExcavator.setName("Terex");
+        secondExcavator.setBucketVolume(2.91);
+
+        Excavator thirdExcavator = new Excavator();
+        thirdExcavator.setColor("blue");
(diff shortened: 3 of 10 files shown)


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,13 +1,15 @@
...
 public class Bulldozer extends Machine {
+    private String bladeType;
+
     public Bulldozer() {
     }
 
+    public void setBladeType(String bladeType) {
+        this.bladeType = bladeType;
+    }
+
```
ZVIIR, line 12: Recommend to create getters and for other similar classes too.

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,32 @@
...
+
+        Bulldozer thirdBulldozer = new Bulldozer();
+        thirdBulldozer.setColor("black");
+        thirdBulldozer.setName("Shantui");
+        thirdBulldozer.setBladeType("S-U blade");
+
+        ArrayList<Bulldozer> bulldozers = new ArrayList<>();
+        bulldozers.add(firstBulldozer);
+        bulldozers.add(secondBulldozer);
+        bulldozers.add(thirdBulldozer);
```
ZVIIR, line 29: I think you don't need to initialize all fields, just create some empty objects in "for" loop.

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,12 @@
...
-    void fill(List<Object> machines, Object value);
-
-    /**
-     * Call the method doWork() from each machine.
-     * This method should be able to accept a list of bulldozers as well as list of trucks.
-     *
-     * @param machines - the list of machines
-     */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
ZVIIR, line 11: You can extends or super from T type, it extends Machine already.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+            BulldozerProducer<T> bulldozerProducer = new BulldozerProducer<>();
+            return (List<T>) bulldozerProducer.get();
+        } else if (type.equals(Excavator.class)) {
+            ExcavatorProducer<T> excavatorProducer = new ExcavatorProducer<>();
+            return (List<T>) excavatorProducer.get();
+        } else if (type.equals(Truck.class)) {
+            TruckProducer<T> truckProducer = new TruckProducer<>();
+            return (List<T>) truckProducer.get();
+        }
+        return new ArrayList<>();
```
ZVIIR, line 23: https://mate-academy.github.io/jv-program-common-mistakes/java-core/pecs/pecs.html#consider-creating-a-local-variable-in-your-implementation-of-getallclass-type-method

ZVIIR (changes requested):
Great solution! I recommend making some changes in your code, according to common mistakes, and for better readability.

src/main/java/core/mate/academy/model/Bulldozer.java
Someboty: They didn't say "always create getters". They said "You should never want to expose the object fields directly. They should be accessed through special methods", so if you need read it, you should create getter. But i don't

src/main/java/core/mate/academy/service/BulldozerProducer.java
Someboty: Oh. I thought they are checking if objects are not null. Ok, I'll remove it

src/main/java/core/mate/academy/service/MachineServiceImpl.java
Someboty: My bad

src/main/java/core/mate/academy/service/MachineService.java
Someboty: I got "Unexpected bound" compilation error, if I doing this

Someboty: Oh, I've got it

ZVIIR (approved):
LGTM!

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer<E> implements MachineProducer<E> {
```
andrii-hoienko, line 7: Parametrization is not used, please see common mistakes (https://mate-academy.github.io/jv-program-common-mistakes/java-core/pecs/pecs.html#do-not-use-raw-types-always-parametrize-interfaces-and-classes-if-it-is-possible) This class should work only with excavator

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,10 @@
...
 import java.util.List;
 
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<E> {
+    int LIST_SIZE = 3;
```
andrii-hoienko, line 7: don`t forget about access modifiers is package private is the best option here? Also rename this to represent its sence like default number of machines

andrii-hoienko (changes requested):
Good, but let`s improve your solution

src/main/java/core/mate/academy/service/MachineProducer.java
Someboty: It is not package protected, it is public static final and I cannot change its access parameters. Did you mean that I need to create a private static final variable separately in each class? Although they will be a duplication in this case and do not fit the  DRY principle.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+    @Override
+    public List<T> getAll(Class<? extends T> type) {
+        if (type.equals(Bulldozer.class)) {
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
```
nacenik, line 14: you can create class variable and reuse it (apply for all producers)

nacenik (approved):
LGTM! 👍

PR #657, opened 2023-07-17
Jv pecs solution

keda-vlad:


src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,15 @@
...
+public class TruckProducer implements MachineProducer<Truck> {
+    private static final Truck first = new Truck();
+    private static final Truck second = new Truck();
+    private static final Truck third = new Truck();
+
+    @Override
+    public List<Truck> get() {
+        return List.of(first, second, third);
+    }
+}
```
holsoni, line 15: You create three same equal constants,  i think it's redundant. You can try this solution.
```suggestion
    private static final int SIZE_OF_LIST = 3;

    @Override
    public List<Truck> get() {
        List<Truck> list = new ArrayList<>();
        for (int i = 0; i < SIZE_OF_LIST; i++) {
             list.add(new Truck());
        }
        return list;
    }
}
```

holsoni (changes requested):
Great solution! I also reccomend to add some fields in classes Truck, BUlldozer and Excavator, because it was stated in comments before class declaration.

src/main/java/core/mate/academy/service/TruckProducer.java
keda-vlad: I dont think its need this because it is template, a was lazy to add some fields in those classes, and if i did it i couldn't use cycle))

holsoni (approved):

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private static final Bulldozer first = new Bulldozer();
```
andrii-hoienko, line 7: Please use loop to create objects (see common mistakes lottery task)

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    private static final Excavator first = new Excavator();
```
andrii-hoienko, line 7: Please use loop to create objects (see common mistakes lottery task)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,47 @@
...
-public class MachineServiceImpl {
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+    private BulldozerProducer bulldozerProducer;
```
andrii-hoienko, line 11: Shoud be final as well, and cound be initialized on field level

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,47 @@
...
+    public List<T> getAll(Class<? extends T> type) {
+        List<? extends Machine> machines = List.of();
+        if (type.equals(Bulldozer.class)) {
+            machines = bulldozerProducer.get();
+        } else if (type.equals(Excavator.class)) {
+            machines = excavatorProducer.get();
+        } else if (type.equals(Truck.class)) {
+            machines = truckProducer.get();
+        }
+        return new ArrayList<>((List<T>) machines);
```
andrii-hoienko, line 31: What is the reason to create machines if you create new ArrayList anyway

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,47 @@
...
+
+    public MachineServiceImpl() {
+        bulldozerProducer = new BulldozerProducer();
+        excavatorProducer = new ExcavatorProducer();
+        truckProducer = new TruckProducer();
+    }
+
+    @Override
+    public List<T> getAll(Class<? extends T> type) {
+        List<? extends Machine> machines = List.of();
```
andrii-hoienko, line 23: better to new ArrayList instead of List.of()

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Truck;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+    private static final Truck first = new Truck();
```
andrii-hoienko, line 7: Pleae use loop

andrii-hoienko (changes requested):
Good, but let`s improve your solution

src/main/java/core/mate/academy/service/MachineServiceImpl.java
keda-vlad: I use this to return an empty List if the class fails all checks

nacenik: It doesn't matter.   `new ArrayList` will create a modified list, that better then List.of()

nacenik (approved):
LGTM! 👍

PR #658, opened 2023-07-19
Created solution for task jv-pecs

Stelmaschyk:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,21 @@
...
+    private List<Bulldozer> bulldozers = List.of(bulldozer1, bulldozer2, bulldozer3);
+
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> listOfBulldozers = new ArrayList<Bulldozer>();
+        for (Object element : bulldozers) {
+            listOfBulldozers.add((Bulldozer) element);
+        }
+        return listOfBulldozers;
+    }
```
okuzan, line 20:
```suggestion
    @Override
    public List<Bulldozer> get() {
        return List.of(new Bulldozer(), new Bulldozer(), new Bulldozer());
    }
```
if you opt to use List.of, we can do it way simpler

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public List<Truck> get() {
+        List<Truck> listOfTrucks = new ArrayList<>();
+        for (Object element : trucks) {
+            listOfTrucks.add((Truck) element);
+        }
+        return listOfTrucks;
+    }
+}
+
```
okuzan, line 22: line

okuzan (changes requested):

olekskov (approved):
Good job

PR #660, opened 2023-07-22
implements methods

LiudmylaPechevysta:


src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -1,6 +1,6 @@
 package core.mate.academy.model;
 
-public abstract class Machine implements Workable {
+public abstract class Machine<T> implements Workable {
```
Ivan95kos, line 3: it is unnecessary to do this
```suggestion
public abstract class Machine implements Workable {
```

src/main/java/core/mate/academy/model/Workable.java
```diff
@@ -1,6 +1,6 @@
 package core.mate.academy.model;
 
-public interface Workable {
+public interface Workable<T extends Machine> {
```
Ivan95kos, line 3: also
```suggestion
public interface Workable {
```

Ivan95kos (approved):

PR #664, opened 2023-07-26
pecs hw is done

GiantBear321:
I want to ask, is there anywhere some info about connection of interface method with generics and PECS and his implementation? I tried to experiment a bit with it, but still loads of questions. Like why should i bother writing some PECS and generics in interfaces if i still have to write the same in the implementation? I think this homework is like too... big, too many details, it gave me global understanding, but left white spaces in details... What should i do? Continue surfing the Internet, go for QnA or let it be, because i will get details later on java core?

### DIFF
--- a/src/main/java/core/mate/academy/model/Excavator.java
+++ b/src/main/java/core/mate/academy/model/Excavator.java
@@ -5,6 +5,14 @@ package core.mate.academy.model;
  * Do not remove no-args constructor
  */
 public class Excavator extends Machine {
+    private String color;
+    private int year;
+
+    public Excavator(String color, int year) {
+        this.color = color;
+        this.year = year;
+    }
+
     public Excavator() {
     }
 
--- a/src/main/java/core/mate/academy/service/BulldozerProducer.java
+++ b/src/main/java/core/mate/academy/service/BulldozerProducer.java
@@ -1,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer{
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        bulldozers.add(new Bulldozer("Red", 2001));
+        bulldozers.add(new Bulldozer("Green", 2004));
+        return bulldozers;
+    }
+}
--- a/src/main/java/core/mate/academy/service/ExcavatorProducer.java
+++ b/src/main/java/core/mate/academy/service/ExcavatorProducer.java
@@ -1,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer{
+    @Override
+    public List<Excavator> get() {
+        List<Excavator> excavators = new ArrayList<>();
+        excavators.add(new Excavator("yellow", 1998));
+        excavators.add(new Excavator("black", 2000));
+        return excavators;
+    }
+}
--- a/src/main/java/core/mate/academy/service/MachineProducer.java
+++ b/src/main/java/core/mate/academy/service/MachineProducer.java
@@ -1,11 +1,9 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
+
 import java.util.List;
 
-public interface MachineProducer {
-    /**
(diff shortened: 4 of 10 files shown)


sokimaaa:
> I want to ask, is there anywhere some info about connection of interface method with generics and PECS and his implementation? I tried to experiment a bit with it, but still loads of questions. Like why should i bother writing some PECS and generics in interfaces if i still have to write the same in the implementation? I think this homework is like too... big, too many details, it gave me global understanding, but left white spaces in details... What should i do? Continue surfing the Internet, go for QnA or let it be, because i will get details later on java core?

If you still have questions, I would suggest visiting QnA and ask mentor directly what you are struggling with.

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,8 @@
...
 import java.util.List;
 
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<T extends Machine> {
+    List<? extends Machine> get();
```
sokimaaa, line 7: use `T` instead of `? extends Machine`.
```
List<T> get();
```

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,6 +5,14 @@
  * Do not remove no-args constructor
  */
 public class Excavator extends Machine {
+    private String color;
```
sokimaaa, line 8: haven't you already added these fields in `abstract class Machine`?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+
+    @Override
+    public void startWorking(List<? extends T> machines) {
+        for (T machine: machines) {
+            machine.doWork();
+        }
+    }
+
+    @Override
+    public void fill(List machines, Machine value) {
```
sokimaaa, line 41: dont use raw types.

See https://mate-academy.github.io/jv-program-common-mistakes/java-core/pecs/pecs.html

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+        } else if (type == Excavator.class) {
+            return (List<T>) new ArrayList<>(excavatorMachineProducer.get());
+        } else if (type == Truck.class) {
+            return (List<T>) new ArrayList<>(truckMachineProducer.get());
+        }
+        return new ArrayList<>();
+    }
+
+    @Override
+    public void startWorking(List<? extends T> machines) {
```
sokimaaa, line 34: you can simplify field params like 
```
void startWorking(List<T> machines)
```

Because T already represents the parametrized type of Machine.

sokimaaa (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
GiantBear321: But it doesnt work, because program will ask for type "Machine" and types inheritors won`t compile, so i wrote following PECS rules: "List<? extend Machine>", because inheritors are producers for method. Is it ok for home-work?

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer {
```
aaananas, line 7: you should parametrize it

src/main/java/core/mate/academy/service/MachineServiceImpl.java
aaananas: You should not ask questions here, mentor cant see them here, write to jv_chat

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer {
```
aaananas, line 7: parametrize

aaananas (changes requested):

aaananas (approved):

PR #667, opened 2023-07-28
Implemented methods get,getAll,fill,startWorking.

opopovych:
Added classes BuldozerProducer,ExcavatorProducer,TruckProducer.

### DIFF
--- a/src/main/java/core/mate/academy/service/ExcavatorProducer.java
+++ b/src/main/java/core/mate/academy/service/ExcavatorProducer.java
@@ -1,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    @Override
+    public List<Machine> get() {
+        List<Machine> excavatorList = new ArrayList<>();
+        excavatorList.add(new Excavator());
+        return excavatorList;
+    }
+}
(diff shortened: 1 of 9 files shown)


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,13 @@
...
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
+     *
      * @return - the list of models
      */
-    List<Object> get();
+    List<Machine> get();
```
sokimaaa, line 12: you can parametrize the output of the method as well, e.g.
```
List<T> get();
```

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,15 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    @Override
+    public List<Machine> get() {
```
sokimaaa, line 10: I think ExcavatorProducer should return a list of Excavators, doesn't it? 

Otherwise, a possible situation that ExcavatorProducer#get may return a Bulldozer, that isn't expected logic.

sokimaaa (changes requested):

aaananas (approved):

PR #671, opened 2023-08-05
jv-pecs-solution

literandltx:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +14,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
olekskov, line 17: Parametrization should be used in all methods in order to make interface more flexible
```suggestion
    List<T> getAll(Class<? extends T> type);
```

olekskov (approved):
Good job, approved
Please pay attention to comment below for further

PR #682, opened 2023-08-19
implemented machine service producer

RedArtelerist:


src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -4,6 +4,15 @@ public abstract class Machine implements Workable {
     private String name;
     private String color;
 
+    public Machine() {
+
```
ystankevych, line 8: Please, delete empty line in default Constructor
```suggestion
public Machine() {} 

or public Machine() {
}
```

ystankevych (approved):
Great solution. Maxim, one minor comment. I approved but, please, make fixes.

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Random;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    private static final int MAX_NUMBER = 10;
+    private static final int MIN_NUMBER = 1;
+    private final Random random = new Random();
+
+    @Override
+    public List<Excavator> get() {
+        int number = random.nextInt(MAX_NUMBER) + MIN_NUMBER;
```
NazarSmith228, line 15: there is no sense to use two numbers (min and max)

you could do (max number + 1)

however do you really need `Random` here? if  I were you I would simply use my max number in `for` loop :)

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class type);
```
NazarSmith228, line 20: you shouldn't use raw type here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    private final BulldozerProducer bulldozerProducer = new BulldozerProducer();
+    private final TruckProducer truckProducer = new TruckProducer();
+    private final ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+
+    @Override
+    public List<Machine> getAll(Class type) {
```
NazarSmith228, line 20: you shouldn't use raw type here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+        List<? extends Machine> machines = null;
+        if (type == Bulldozer.class) {
```
NazarSmith228, line 22: hint: you can use `isAssignableFrom()` method

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    private final BulldozerProducer bulldozerProducer = new BulldozerProducer();
+    private final TruckProducer truckProducer = new TruckProducer();
+    private final ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+
+    @Override
+    public List<Machine> getAll(Class type) {
+        List<? extends Machine> machines = null;
```
NazarSmith228, line 21: how about assigning `machines` to `List.of()` or `Collections.emptyList()` immediately and not doing that null check at the end?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+        }
+        if (type == Truck.class) {
+            machines = truckProducer.get();
+        }
+        return machines == null ? List.of() : new ArrayList<>(machines);
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        Collections.fill(machines, value);
```
NazarSmith228, line 36: try not to rely on libraries for now

you need to practice more yourself

NazarSmith228 (changes requested):
Very nice! Let's improve your solution now

src/main/java/core/mate/academy/service/MachineServiceImpl.java
RedArtelerist: Can you explain, why this method differs from general equals operator, because I can't exactly understand? Why is it better than equals?))

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -17,23 +16,25 @@ public class MachineServiceImpl implements MachineService<Machine> {
     private final ExcavatorProducer excavatorProducer = new ExcavatorProducer();
 
     @Override
-    public List<Machine> getAll(Class type) {
-        List<? extends Machine> machines = null;
-        if (type == Bulldozer.class) {
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<? extends Machine> machines = List.of();
+        if (Bulldozer.class.isAssignableFrom(type)) {
```
RedArtelerist, line 21: I still don’t understand why we need `isAssignableFrom` here and how it is better than the usual `equals`((

src/main/java/core/mate/academy/service/MachineServiceImpl.java
NazarSmith228: `==` compares references ("location" in memory) as you know
`isAssignableFrom()` is more complex - it allows you to understand whether right-side class can be assigned to a left-side class 

like `Class1.isAssignableFrom(Class2)` will return true if a variable of type `Class2` can be assigned to a variable of type `Class1`

in our case we indeed can use `==` , since we want to check an exact match

therefore I would consider my previous comment as unnecessary, since I missed that we need an exact match only :)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
NazarSmith228: as I mentioned earlier we indeed can get around with just `==` or `equals` - actually they do the same, since `equals()` in `Class` class (I know sounds weird) isn't overriden from `Object` class

in this particular scenario the difference is next: 

suppose we have `Buldozer` and `BuldozerChild extends Buldozer`:

- if type = `Buldozer` - `isAssignableFrom()` will return true and `==` will return true
- if type = `BuldozerChild` - `isAssignableFrom()` will return true but `==` will return false

hopefully now you understand the difference

and again - usage of `isAssignableFrom()` is not mandatory here, since we care about exact classes match, my previous comment may be considered unnecessary :)

NazarSmith228 (approved):
LGTM! :)

PR #683, opened 2023-08-19
Implemented MachineServiceImpl , MachineProducer classes

ystankevych:


RedArtelerist (approved):
Great use of Strategy pattern, I forgot about it) So your solution is excellent

src/main/java/core/mate/academy/service/MachineProducerStrategy.java
```diff
@@ -0,0 +1,23 @@
...
+        switch (type.getSimpleName()) {
+            case "Bulldozer":
+                return Optional.of(new BulldozerProducerImpl());
+            case "Excavator":
+                return Optional.of(new ExcavatorProducerImpl());
+            case "Truck":
+                return Optional.of(new TruckProducerImpl());
+            default:
+                return Optional.empty();
+        }
```
NazarSmith228, line 20: `Optional` is usually used when you know that something is nullable

in this case usage of `Optional` is unjustified - you may add some default implementation of `MachineProducer` for that purpose or return one from existing

src/main/java/core/mate/academy/service/producerimpl/BulldozerProducerImpl.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.IntStream;
+
+public class BulldozerProducerImpl implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        IntStream.range(1, 10)
+                .forEach(i -> bulldozers.add(new Bulldozer()));
```
NazarSmith228, line 14: you added a custom property to `Buldozer` but you don't initialize it here - consider declaring another constructor

src/main/java/core/mate/academy/service/producerimpl/ExcavatorProducerImpl.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.IntStream;
+
+public class ExcavatorProducerImpl implements MachineProducer<Excavator> {
+    @Override
+    public List<Excavator> get() {
+        List<Excavator> excavators = new ArrayList<>();
+        IntStream.range(1, 10)
+                .forEach(i -> excavators.add(new Excavator()));
```
NazarSmith228, line 14: you added a custom property to `Excavator` but you don't initialize it here - consider declaring another constructor

src/main/java/core/mate/academy/service/producerimpl/TruckProducerImpl.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.IntStream;
+
+public class TruckProducerImpl implements MachineProducer<Truck> {
+    @Override
+    public List<Truck> get() {
+        List<Truck> trucks = new ArrayList<>();
+        IntStream.range(1, 10)
+                .forEach(i -> trucks.add(new Truck()));
```
NazarSmith228, line 14: you added a custom property to `Truck` but you don't initialize it here - consider declaring another constructor

src/main/java/core/mate/academy/service/producerimpl/BulldozerProducerImpl.java
```diff
@@ -0,0 +1,17 @@
...
+import core.mate.academy.service.MachineProducer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.IntStream;
+
+public class BulldozerProducerImpl implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        IntStream.range(1, 10)
```
NazarSmith228, line 13: the usage of `IntStream` for simple iteration is unjustified

comparing to a simple `for` loop, stream requires much more resources and gives no benefits in this case

**please apply the same to other implementations**

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,36 @@
...
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        return producerStrategy.getMachineProducerStrategy(type)
+                .map(machineProducer -> new ArrayList<Machine>(machineProducer.get()))
+                .orElseGet(ArrayList::new);
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        if (machines != null) {
```
NazarSmith228, line 27: is this check necessary?

src/main/java/core/mate/academy/service/producerimpl/ExcavatorProducerImpl.java
```diff
@@ -0,0 +1,17 @@
...
+import core.mate.academy.service.MachineProducer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.IntStream;
+
+public class ExcavatorProducerImpl implements MachineProducer<Excavator> {
+    @Override
+    public List<Excavator> get() {
+        List<Excavator> excavators = new ArrayList<>();
+        IntStream.range(1, 10)
```
NazarSmith228, line 13: think of magic numbers :)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,36 @@
...
+            Collections.fill(machines, value);
```
NazarSmith228, line 28: try not to rely on libraries for now - you need to practice more yourself

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,36 @@
...
+    private final MachineProducerStrategy producerStrategy;
+
+    public MachineServiceImpl() {
+        producerStrategy = new MachineProducerStrategy();
+    }
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        return producerStrategy.getMachineProducerStrategy(type)
+                .map(machineProducer -> new ArrayList<Machine>(machineProducer.get()))
```
NazarSmith228, line 21: consider creating a separate variable for `machineProducer.get()` (taking into account that `Optional` will be removed)

NazarSmith228 (changes requested):
Great job! Let's improve your solution :)

mmosii (approved):

src/main/java/core/mate/academy/service/MachineProducerStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+            Class<? extends Machine> type) {
+        switch (type.getSimpleName()) {
+            case "Bulldozer":
+                return new BulldozerProducerImpl();
+            case "Excavator":
+                return new ExcavatorProducerImpl();
+            case "Truck":
+                return new TruckProducerImpl();
+            default:
+                return (MachineProducer<Machine>) ArrayList::new;
```
NazarSmith228, line 20: okay, so what will happen when you approach `default` branch?

`MachineProducer<Machine>` is definitely not a `Supplier<ArrayList>` so I suppose you'll get `ClassCastException`

how about returning one of the existing producers?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
NazarSmith228: not addressed :)

src/main/java/core/mate/academy/service/MachineProducerStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.service.producerimpl.BulldozerProducerImpl;
+import core.mate.academy.service.producerimpl.ExcavatorProducerImpl;
+import core.mate.academy.service.producerimpl.TruckProducerImpl;
+import java.util.ArrayList;
+
+public class MachineProducerStrategy {
+    public MachineProducer<? extends Machine> getMachineProducerStrategy(
+            Class<? extends Machine> type) {
+        switch (type.getSimpleName()) {
```
NazarSmith228, line 12: since you're operating on type, it's weird to use class name

why not comparing classes directly i.e. `type == Buldozer.class`?

NazarSmith228 (changes requested):
One last step :)

src/main/java/core/mate/academy/service/MachineProducerStrategy.java
ystankevych: I can't use type in switch case. Ok, I changed it ti if else and used type==

src/main/java/core/mate/academy/service/MachineProducerStrategy.java
ystankevych: This is the method refference and we have some check for wrong type - all tests passed. Ok, I changed it to full anonymous implementation ). 
"how about returning one of the existing producers?" We have some limits on this in Common mistakes.

ystankevych: and really don't understand why I have to create default Machine producer implementation with method get wchich return empty list. In this case it is really better to use some anonymous impl (method refference or lambda) and return empty list. Ok, but if you really consist on it I will create some class like 
class MachineProducerimpl{
@Override
 method get() {
return empty list;
}
}

=)

NazarSmith228: oh wait...

seems like I'm wrong here :) 

`ClassCastException` will not happen, since ArrayList::new lambda suits to MachineProducer (cause it's a functional interface)

you've implemented a nice trick,  which I didn't even notice at the first time

great job!)

src/main/java/core/mate/academy/service/MachineProducerStrategy.java
NazarSmith228: and you weren't suppose to use switch :)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
NazarSmith228: not addressed x2 :)

pay attention to this remark :D

NazarSmith228 (changes requested):
I would have already approved it, however you forgot about the remark with null check again :)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,35 @@
...
+public class MachineServiceImpl implements MachineService<Machine> {
+    private final MachineProducerStrategy producerStrategy;
+
+    public MachineServiceImpl() {
+        producerStrategy = new MachineProducerStrategy();
+    }
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        return new ArrayList<>(producerStrategy.getMachineProducerStrategy(type).get());
```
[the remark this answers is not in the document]
ystankevych: About not addressed - You mean, NPE is better than checking for null?(If null is passed as argument) Ok, maybe it's because of  a new concept or material in our course that's a little outdated.. I will raise this question. Fixed.
And it was like an eye opening for me...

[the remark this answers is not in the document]
ystankevych: public List<Machine> getAll(Class<? extends Machine> type) {
        MachineProducer<? extends Machine> producer = producerStrategy
                .getMachineProducerStrategy(type);
        return new ArrayList<>(producer.get());
    }

and I leave comment about this peace of code. I can't initialize producer in one line because if I do, the line length will be over  100 characters and clean/package check will fail.

liliia-ponomarenko (approved):
Good job! 👏

PR #685, opened 2023-08-23
Implemented MachineServiceImpl class

mmosii:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
```
teract10s, line 18: Meybe it's wil be better to have this variable in class, like constant. But it's up to you

teract10s (approved):

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        ArrayList<Bulldozer> machines = new ArrayList<>();
+        machines.add(new Bulldozer());
+        machines.add(new Bulldozer());
+        machines.add(new Bulldozer());
```
NazarSmith228, line 13: why not using a `for` loop with a predefined constant?

**should be applied to all other implementations**

src/main/java/core/mate/academy/service/MachineServiceImpl.java
NazarSmith228: agree - I would create 3 `private final` fields in this class - one for each producer

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
```
NazarSmith228, line 17: how about initializing and empty list at the first line

then check the type via `if` statements and reassign that list to `producer.get()`

it will make your code much simpler and remove all the unnecessary `return` statements`

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+        if (type == Excavator.class) {
+            ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+            return new ArrayList<>(excavatorProducer.get());
+        }
+        return new ArrayList<>();
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
+        int size = machines.size();
```
NazarSmith228, line 34: why not inline this in the loop - `i < machines.size()`

NazarSmith228 (changes requested):
Great job! Now let's improve your solution

Elena-Bruyako (approved):

PR #692, opened 2023-08-26
created MachineServiceImpl class + all related classes & interfaces

validatorr:


torbinofficial (approved):
Everything looks okay, but delete all unnecessary comments

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,10 @@
...
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<T extends Machine> {
+    int AMOUNT_OF_MODELS = 3;
+
+    List<? extends Machine> get();
```
NazarSmith228, line 9: you should not use `Machine` explicitly - you have a parameterized type

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
-     * Return the list of machines.
-     * In the implementation of this method please use your MachineProducer implementations
-     * See that 'Class type' is not parametrized.
-     * Consider to parametrize this Class< PARAMETRIZE_ME > with specific wildcard bounding
-     *
-     * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
-     * @return the list of machines
-     */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class<? extends Machine> type);
```
NazarSmith228, line 8: you should not use `Machine` explicitly - you have a parameterized type

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
-    /**
-     * Fill the machines list with passed value
-     * Replace the Object with parametrized value.
-     * This method should be able to work well with any type of machine passed as 'value'
-     *
-     * @param machines - list of machines to be filled with value
-     * @param value    - any object of machine sub class
-     */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, T value);
```
NazarSmith228, line 10: you should not use `Machine` explicitly - you have a parameterized type

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
 
-    /**
-     * Call the method doWork() from each machine.
-     * This method should be able to accept a list of bulldozers as well as list of trucks.
-     *
-     * @param machines - the list of machines
-     */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
NazarSmith228, line 12: you should not use `Machine` explicitly - you have a parameterized type

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(Bulldozer.class)) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if (type.equals(Excavator.class)) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else if (type.equals(Truck.class)) {
+            return new ArrayList<>(new TruckProducer().get());
+        } else {
+            return Collections.emptyList();
+        }
```
NazarSmith228, line 23: this can be simplified

you can init the returning list as empty and then just reassign for a corresponding producer

if no producer found - just return the original empty list

this way you'll have just 1 `return`, 1 `if`, 1 `else if` and `else` statements

and there is no sense to wrap `producer.get()` into additonal `ArrayList` - producers already return `List`

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+    private final BulldozerProducer bulldozerProducer = new BulldozerProducer();
+    private final ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+    private final TruckProducer truckProducer = new TruckProducer();
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
```
Elena-Bruyako, line 17: You can create new ArrayList here and assign list for each type so you avoid return statement in every checking

src/main/java/core/mate/academy/service/MachineProducer.java
Elena-Bruyako: need to be fixed

Elena-Bruyako (changes requested):
Good job! Let's fix some comments

PR #693, opened 2023-08-27
implemented three more classes with overriding get() and implemented …

vsych321:
…MachineServiceImpl class

### DIFF
--- a/src/main/java/core/mate/academy/model/Machine.java
+++ b/src/main/java/core/mate/academy/model/Machine.java
@@ -3,6 +3,8 @@ package core.mate.academy.model;
 public abstract class Machine implements Workable {
     private String name;
     private String color;
+    public Machine() {
+    }
 
     public String getName() {
         return name;
--- a/src/main/java/core/mate/academy/service/MachineServiceImpl.java
+++ b/src/main/java/core/mate/academy/service/MachineServiceImpl.java
@@ -1,7 +1,44 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import core.mate.academy.specificImp.BulldozerProducer;
+import core.mate.academy.specificImp.ExcavatorProducer;
+import core.mate.academy.specificImp.TruckProducer;
+
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine>{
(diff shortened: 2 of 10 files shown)


src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -4,6 +4,9 @@ public abstract class Machine implements Workable {
     private String name;
     private String color;
 
+    public Machine() {
+    }
+
```
NazarSmith228, line 9: why do you need this?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<? extends Machine> machines = new ArrayList<>();
+        if (type.equals(Bulldozer.class)) {
+            machines = new BulldozerProducer().get();
+        } else if (type.equals(Excavator.class)) {
+            machines = new ExcavatorProducer().get();
+        } else if (type.equals(Truck.class)) {
+            machines = new TruckProducer().get();
```
NazarSmith228, line 25: consider creating producers beforehand - as final fields, so that you don't create a new instance each time this method is called

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        }
+        return (List<Machine>) machines;
```
NazarSmith228, line 27: you should not introduce casting here

src/main/java/core/mate/academy/specificimp/BulldozerProducer.java
```diff
@@ -0,0 +1,19 @@
+package core.mate.academy.specificimp;
+
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.service.MachineProducer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private static final int SPECIFIC_LIST = 10;
```
NazarSmith228, line 9: weird contant name - how about `NUMBER_OF_BULLDOZERS`?

**should be applied to all other producers**

NazarSmith228 (changes requested):
Good job! Let's improve your solution

src/main/java/core/mate/academy/service/MachineServiceImpl.java
vsych321: but it won't work without casting

liliia-ponomarenko (approved):
Great 😊

PR #701, opened 2023-09-06
Implemented the following methods as per the task requirements: getAl…

Petro-Smoliar:
…l(Class type), fill(List<Object> machines, Object value), and startWorking()


Ivan95kos (approved):
Great job

okuzan (approved):

PR #707, opened 2023-09-18
add MachineService & MacineProducer implementations

VoitseshukV:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
olekskov, line 9: We should limit possible types in order not to give user create MachineService parametrized with Fruit, Animal etc
```suggestion
public interface MachineService<T extends Machine> {
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
...
+    void fill(List<? super T> machines, T value);
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
olekskov, line 38: We should base method on main parameter
```suggestion
    void startWorking(List<? extends T> machines);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T> implements MachineService {
```
olekskov, line 13:
```suggestion
public class MachineServiceImpl implements MachineService<Machine> {
```

olekskov (changes requested):
Need to fix parametrization in some cases.
Methods of MachineServiceImpl probably will require fixes to work

olekskov (approved):

PR #711, opened 2023-10-02
Realized producers for all machine types. In realization used PECS.

Trayvor:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class type);
```
kozhukhovsky, line 20:
```suggestion
    List<T> getAll(Class<...> type);
```

add paramater instead ...

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,62 @@
...
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
```
kozhukhovsky, line 14:
```suggestion
```

kozhukhovsky (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,61 @@
...
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(Truck.class)) {
```
okuzan, line 16: When comparing Class objects in Java, it's more typical to use == rather than equals. This is because there's only one Class object for each type, so you can safely use identity comparison (==).

okuzan (changes requested):

okuzan (approved):
Well done!

PR #712, opened 2023-10-03
implemented machineService using PECS

dima-semeniuk:


flekros (changes requested):
Great job!

flekros (approved):
Great job!

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+    private static final int TRUCK_COUNT = 3;
+
+    @Override
+    public List<Truck> get() {
+        List<Truck> listOfTrucks = new ArrayList<>();
+        for (int i = 0; i < TRUCK_COUNT; i++) {
+            listOfTrucks.add(new Truck());
```
kshuryhin, line 14: We probably want to create our trucks with some data, so consider adding all-args constructor for those purposes. Same about other models

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,9 +5,20 @@
  * Do not remove no-args constructor
  */
 public class Excavator extends Machine {
+    private String typeOfFuel;
```
kshuryhin, line 8: It looks like we are not using `typeOfFuel` variable at all

kshuryhin (changes requested):
Good job, but consider some improvements related to your models

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<? extends Machine> machines = new ArrayList<>();
+        if (type == Bulldozer.class) {
+            machines = new BulldozerProducer().get();
+        }
```
kshuryhin, line 19: One more minor point, probably we want to use here _if-else-if_, because we gonna have only one type, so no need for all checks, if one was already found

kshuryhin (changes requested):
One minor point to resolve

kshuryhin (approved):
Great!

PR #718, opened 2023-10-10
Add implementation for the PECS task

MinoDentori:


bhdnchui (changes requested):
Great solution!
But a few comments need to be processed.

kshuryhin (approved):
Good job!

PR #719, opened 2023-10-11
Jv pecs solution

SosunovychB:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,10 @@
...
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<T extends Machine> {
+    int NUMBER_OF_MACHINES = 3;
+
+    List<? extends Machine> get();
```
Rommelua, line 9: What is the reason for having this **T** if we do not use it?
image

Rommelua (changes requested):

src/main/java/core/mate/academy/service/MachineProducer.java
SosunovychB: The logic of using of this restricted generic is that interface will also be used for Machine subclasses. Simulataneously, typical generics are invariant.

sokimaaa, line 9: use generic type that already bounded. 
```suggestion
    List<T> get();
```

sokimaaa (changes requested):

Rommelua (approved):

PR #721, opened 2023-10-14
realized MachineServiceimpl class

fedorovychh:
added producers classes

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineServiceImpl.java
+++ b/src/main/java/core/mate/academy/service/MachineServiceImpl.java
@@ -1,7 +1,39 @@
 package core.mate.academy.service;
 
-/**
- * Your implementation of MachineService.
- */
-public class MachineServiceImpl {
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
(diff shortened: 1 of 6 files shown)


DevIgork (approved):
good job!

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(Bulldozer.class)) {
+            machines = new BulldozerProducer().get();
+        } else if (type.equals(Excavator.class)) {
+            machines = new ExcavatorProducer().get();
+        } else if (type.equals(Truck.class)) {
+            machines = new TruckProducer().get();
+        } else {
+            return new ArrayList<>();
+        }
```
kshuryhin, line 24: Probably we can omit this case, as if type is none of the above, then `machines` list will be empty anyway.

kshuryhin (changes requested):
Good job, but please, consider adding some specific fields per each subtype of the Machine class.

PR #732, opened 2023-10-29
implemented task

yreuta:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            List<? extends Machine> machines = new BuldozerProducer().get();
+            return new ArrayList<>(machines);
```
kerrrusha, line 19: no need to create new `Producer` each time you want to call `.get()` method on it, make this producer as a class `private final` field 
```suggestion
           return new ArrayList<>(bulldozerProducer.get());
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        } else if (type == Excavator.class) {
+            List<? extends Machine> machines = new ExcavatorProducer().get();
+            return new ArrayList<>(machines);
```
kerrrusha, line 22: no need to create new `Producer` each time you want to call `.get()` method on it, make this producer as a class `private final` field 
```suggestion
           return new ArrayList<>(excavatorProducer.get());
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        } else if (type == Truck.class) {
+            List<? extends Machine> machines = new TruckProducer().get();
+            return new ArrayList<>(machines);
```
kerrrusha, line 25: no need to create new `Producer` each time you want to call `.get()` method on it, make this producer as a class `private final` field 
```suggestion
           return new ArrayList<>(truckProducer.get());
```

kerrrusha (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            List<? extends Machine> machines = buldozerProducer.get();
+            return new ArrayList<>(machines);
+        } else if (type == Excavator.class) {
+            List<? extends Machine> machines = excavatorProducer.get();
+            return new ArrayList<>(machines);
+        } else if (type == Truck.class) {
+            List<? extends Machine> machines = truckProducer.get();
+            return new ArrayList<>(machines);
```
kerrrusha, line 28: i've mentioned in previous comment also that these variables is redundant
```suggestion
            return new ArrayList<>(buldozerProducer.get());
        } else if (type == Excavator.class) {
            return new ArrayList<>(excavatorProducer.get());
        } else if (type == Truck.class) {
            return new ArrayList<>(truckProducer.get());
```

kerrrusha (approved):
see comment

PR #733, opened 2023-10-31
Hw-pecs

nastasiiaKulinich:


okuzan (changes requested):
Good job overall! Fix everywhere as per comments for Truck

olekskov (approved):

PR #734, opened 2023-11-01
hm java-pecs

Paradoxallist:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<Object> machines, T value);
```
kerrrusha, line 30: we should use wildcards, not raw `Object`

kerrrusha (changes requested):

kerrrusha (approved):

PR #738, opened 2023-11-05
JV-PECS

YanaBN:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+        if (Bulldozer.class.equals(type)) {
+            List<? extends Machine> machines = new BulldozerProducer().get();
+            return new ArrayList<>(machines);
+        } else if (Excavator.class.equals(type)) {
+            List<? extends Machine> machines = new ExcavatorProducer().get();
+            return new ArrayList<>(machines);
+        } else if (Truck.class.equals(type)) {
+            List<? extends Machine> machines = new TruckProducer().get();
+            return new ArrayList<>(machines);
+        }
```
kerrrusha, line 26: producers should be `private final` class-level fields
```suggestion
        if (Bulldozer.class.equals(type)) {
            return new ArrayList<>(bulldozerProducer.get());
        } else if (Excavator.class.equals(type)) {
            return new ArrayList<>(excavatorProducer.get());
        } else if (Truck.class.equals(type)) {
            return new ArrayList<>(truckProducer.get());
        }
```

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,11 +5,21 @@
...
         System.out.println("Bulldozer started to work");
     }
+
+    public int getBucketSize() {
+        return bucketSize;
+    }
+
+    public void setBucketSize(int bucketSize) {
+        this.bucketSize = bucketSize;
+    }
```
kerrrusha, line 24: constructor-based field initialization is more preffered over setter-based
```suggestion
    private final int bucketSize;

    public Bulldozer(int bucketSize) {
    this.bucketSize = bucketSize;
    }

    @Override
    public void doWork() {
        System.out.println("Bulldozer started to work");
    }

    public int getBucketSize() {
        return bucketSize;
    }
```

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -5,11 +5,21 @@
...
         System.out.println("Truck started to work");
     }
+
+    public int getTankValue() {
+        return tankValue;
+    }
+
+    public void setTankValue(int tankValue) {
+        this.tankValue = tankValue;
+    }
```
kerrrusha, line 24: constructor-based field initialization is more preffered over setter-based

Rommelua (approved):

PR #739, opened 2023-11-06
Done task with PECS

AntonBabychP1T:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,15 +1,23 @@
...
 
+import core.mate.academy.service.MachineProducer;
+import java.util.List;
+
 /**
  * Add some custom fields that could be only in Bulldozer
  * Do not remove no-args constructor
  */
-public class Bulldozer extends Machine {
+public class Bulldozer extends Machine implements MachineProducer {
```
sokimaaa, line 10: Raw type use -> bad practice.

sokimaaa (changes requested):
Raw type using

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,15 +1,23 @@
...
 
     @Override
     public void doWork() {
         System.out.println("Bulldozer started to work");
     }
+
+    @Override
+    public List<Bulldozer> get() {
+        return List.of(new Bulldozer(), new Bulldozer());
+    }
```
kerrrusha, line 22: i dont understand prev mentor's comment, Bulldozer model is not a producer, no need to implement from MachineProducer
instead, you should create separate package `producer`, and create Producers for each model, implemented MachineProducer<?> interface
also, according to the task, you should add some fields to this model
image

```suggestion
public class Bulldozer extends Machine {
    public Bulldozer() {
    }

    @Override
    public void doWork() {
        System.out.println("Bulldozer started to work");
    }
```

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -1,15 +1,23 @@
...
     @Override
     public void doWork() {
         System.out.println("Excavator started to work");
     }
+
+    @Override
+    public List<Excavator> get() {
+        return List.of(new Excavator(), new Excavator());
+    }
 }
```
kerrrusha, line 23: no need to implement from MachineProducer
instead, you should create separate package `producer`, and create Producers for each model, implemented MachineProducer<?> interface
also, according to the task, you should add some fields to this model
image
```suggestion
public class Excavator extends Machine {
    public Excavator() {
    }

    @Override
    public void doWork() {
        System.out.println("Excavator started to work");
    }
}
```

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -1,15 +1,23 @@
...
     @Override
     public void doWork() {
         System.out.println("Truck started to work");
     }
+
+    @Override
+    public List<Truck> get() {
+        return List.of(new Truck(), new Truck());
+    }
 }
```
kerrrusha, line 23: no need to implement from MachineProducer
instead, you should create separate package `producer`, and create Producers for each model, implemented MachineProducer<?> interface
also, according to the task, you should add some fields to this model
image
```suggestion
public class Truck extends Machine {
    public Truck() {
    }

    @Override
    public void doWork() {
        System.out.println("Truck started to work");
    }
}
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+    @Override
+    public List<T> getAll(Class<? extends T> type) {
+        if (type.equals(Bulldozer.class)) {
+            return (List<T>) new Bulldozer().get();
+        }
+        if (type.equals(Excavator.class)) {
+            return (List<T>) new Excavator().get();
+        }
+        if (type.equals(Truck.class)) {
+            return (List<T>) new Truck().get();
```
kerrrusha, line 21: use producers, not models
also, producers should be created as a class-level `private final` field

kerrrusha (changes requested):

PR #749, opened 2023-11-13
Hw pecs

AGuliaiev:


src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class TruckProducer implements MachineProducer<Truck> {
+    @Override
+    public List<Truck> get() {
+
+        Truck model1 = new Truck();
+        Truck model2 = new Truck();
+        Truck model3 = new Truck();
+
+        return List.of(model1, model2, model3);
```
okuzan, line 14:
```suggestion
        return List.of(new Truck(),...
```

okuzan: you don't need to name a variable if you can use it right away, fix everywhere

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Truck;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+    @Override
+    public List<Truck> get() {
+
```
okuzan, line 9: you don't need an empty line after method/class definition

okuzan (approved):

PR #755, opened 2023-11-18
Hw jv pecs

AdamMudrak:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,9 +5,43 @@
...
+        this.maxSpeedKmH = maxSpeedKmH;
+    }
+
+    public boolean isEcoFriendly() {
+        return isEcoFriendly;
+    }
+
+    public void setEcoFriendly(boolean ecoFriendly) {
+        isEcoFriendly = ecoFriendly;
+    }
```
Sanderfuck: For what you need it?

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,39 @@
...
+    void fill(List<? super Machine> machines, Machine value);
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
Sanderfuck: You have Generic what extends Machine

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,57 @@
...
+    public void fill(List<? super Machine> machines, Machine value) {
+        Class<? extends Machine> type = value.getClass();
+        if (type == Bulldozer.class) {
+            machines.replaceAll(ignored -> new Bulldozer());
+        } else if (type == Excavator.class) {
+            machines.replaceAll(ignored -> new Excavator());
+        } else if (type == Truck.class) {
+            for (int i = 0; i < machines.size(); i++) {
+                machines.replaceAll(ignored -> new Truck());
+            }
```
Sanderfuck: You can fill it using loop and use machines.set

Sanderfuck (changes requested):

src/main/java/core/mate/academy/model/Bulldozer.java
AdamMudrak: image
I thought here they wanted me to add access modifiers + accessor methods even in case I don't use them for the task. So I should delete all getters and setters in all models?

src/main/java/core/mate/academy/service/MachineServiceImpl.java
AdamMudrak: And what's bad about my solution?

src/main/java/core/mate/academy/service/MachineService.java
AdamMudrak: So I should use type "T" instead of "? extends Machine" and "Machine" here?

okuzan: if you don't use your type parameter here, why would you have it in the first place? Use ? extends/super T here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
okuzan: If something can be done with less code without decreasing readability - we usually should prefer it

okuzan (changes requested):

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,39 @@
...
      * In the implementation of this method please use your MachineProducer implementations
      * See that 'Class type' is not parametrized.
      * Consider to parametrize this Class< PARAMETRIZE_ME > with specific wildcard bounding
      *
-     * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
+     * @param type - any class of Machine subclass. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
Sanderfuck:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/producers/MachineProducer.java
```diff
@@ -0,0 +1,13 @@
...
+import core.mate.academy.model.Machine;
+import java.util.List;
+
+public interface MachineProducer<T extends Machine> {
+    /**
+     * In your implementations - create 2-3 models, add them to the list and return
+     *
+     * @return - the list of models
+     */
+    List<? super T> get();
```
Sanderfuck:
```suggestion
    List<T> get();
```

Sanderfuck (changes requested):

Sanderfuck (approved):

PR #761, opened 2023-11-28
 Parameterized classes and methods, added realization of methods needed.

SSosedova:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
```
```suggestion
    List<T> getAll(Class<? extends T> type);
```

okuzan (changes requested):

okuzan (approved):

PR #776, opened 2023-12-19
Parametrized interfaces and implemented required classes.

davydov-software:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
```
okuzan, line 9: in each method, replace machine with T

okuzan (changes requested):

Ivan95kos (approved):
Nice

PR #778, opened 2023-12-25
hm for review

vdbond:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,39 @@
...
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        }
+        if (type == Truck.class) {
+            return new ArrayList<>(new TruckProducer().get());
+        }
+        if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        }
+        return List.of();
```
olekskov, line 23: it's better to use Collections.emptyList() when you need empty list as it provides link to existing object of immutable empty list.
List.of() creates new object which is usually not necessary.
Also Collections.emptyList() more properly describes what do you want to return

olekskov (approved):
Good job
Please pay attention to comment below for future

PR #781, opened 2023-12-27
Jv-pecs task solution

mkazarian-1:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,11 +5,32 @@
...
+    }
+
+    public int getWeight() {
+        return weight;
+    }
+
+    public Bulldozer setWeight(int weight) {
+        this.weight = weight;
+        return this;
+    }
```
koval-kn: Do the same for Excavator and Truck classes
1. use constructor-based fields initialization
2. setters should have `void` return type
```suggestion
    public Bulldozer(int enginePower, int weight) {
    this.enginePower = enginePower;
    this.weight = weight;
    }

    @Override
    public void doWork() {
        System.out.println("Bulldozer started to work");
    }

    public int getEnginePower() {
        return enginePower;
    }

    public int getWeight() {
        return weight;
    }
```

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,13 @@
...
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
+     *
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends T> get();
```
koval-kn:
```suggestion
    List<T> get();
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +16,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
```
koval-kn:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+
+        if (Bulldozer.class.equals(type)) {
+            machines = new BulldozerProducer().get();
+        } else if (Truck.class.equals(type)) {
+            machines = new TruckProducer().get();
+        } else if (Excavator.class.equals(type)) {
+            machines = new ExcavatorProducer().get();
+        }
+
+        return new ArrayList<>(machines);
```
koval-kn: producers should be `private final` class-level fields
```suggestion
        if (Bulldozer.class.equals(type)) {
            return bulldozerProducer().get();
        } else if (Truck.class.equals(type)) {
            return truckProducer().get();
        } else if (Excavator.class.equals(type)) {
            return excavatorProducer().get();
        }
```

koval-kn (changes requested):

sokimaaa (approved):

PR #784, opened 2023-12-30
parameterized methods and implementations and added classes

RCherniavskiy:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
...
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends Machine> get();
```
olekskov: Please use parameter from class in method, otherwise parametrizing class have no sense
```suggestion
    List<? extends T> get();
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, T value);
```
olekskov:
```suggestion
    void fill(List<? super T> machines, T value);
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #797, opened 2024-01-28
made three implementations of the MachineProducer interface and 

Ovikt0r:
implementation of the MachineService interface


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,11 +5,47 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+
```
okuzan, line 8: in mate we don't leave an empty line after class definition

okuzan (approved):
Good job

PR #802, opened 2024-02-02
PECS solution

DojiDoe:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,7 +5,18 @@
...
+
     public Bulldozer() {
+
+    }
+
+    public Bulldozer(int enginePower, int height, String name, String color) {
+        this.enginePower = enginePower;
+        this.height = height;
+        this.setName(name);
+        this.setColor(color);
```
koval-kn: we dont need to use `this` keyword here
```suggestion
        setName(name);
        setColor(color);
```

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -5,9 +5,20 @@
...
+
     public Excavator() {
     }
 
+    public Excavator(int weight, int liftCapacity, String name, String color) {
+        this.weight = weight;
+        this.liftCapacity = liftCapacity;
+        this.setName(name);
+        this.setColor(color);
+    }
```
koval-kn: we dont need to use `this` keyword here

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -5,9 +5,19 @@
...
+    private int numberOfCylinders;
+
     public Truck() {
     }
 
+    public Truck(int length, int numberOfCylinders, String name, String color) {
+        this.length = length;
+        this.numberOfCylinders = numberOfCylinders;
+        this.setName(name);
+        this.setColor(color);
```
koval-kn: we dont need to use `this` keyword here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class type) {
+        MachineProducer machineProducer;
+        switch (type.getSimpleName()) {
+            case "Bulldozer":
```
koval-kn: try this. if it will not work, just rewrite with `if (type.equals(Bulldozer.class)) ...`, because using String literals here is bad practice
```suggestion
        switch (type) {
            case Bulldozer.class:
```

koval-kn (changes requested):

koval-kn (approved):

PR #803, opened 2024-02-02
parametrized MachineProducer & MachineService interfaces, implemented…

DanilKi:
… BulldozerProducer, ExcavatorProducer, TruckProducer & MachineServiceImpl classes

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineService.java
+++ b/src/main/java/core/mate/academy/service/MachineService.java
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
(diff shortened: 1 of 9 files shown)


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class<? extends Machine> type);
```
olekskov: We should use our class parameter here as in all other methods
```suggestion
    List<T> getAll(Class<? extends T> type);
```

olekskov (approved):
Good job, approved

PR #805, opened 2024-02-03
14:05

leonidbondarenko69:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,12 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
```
olekskov, line 5: We should limit possibility to make Machine producer parametrized with Cake or Bird
```suggestion
public interface MachineProducer<T extends Machine> {
```

olekskov (approved):
Good job

PR #810, opened 2024-02-13
jv-pecs

OleksiiBashkirov:
created BulldozerProducer class
created ExcavatorProducer class
created TruckProducer class
added specific fields, getters, setters, constructors to Bulldozer, Excavator, Truck classes 
created MachineServiceImpl class

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineService.java
+++ b/src/main/java/core/mate/academy/service/MachineService.java
@@ -7,7 +7,6 @@ import java.util.List;
  */
 
 public interface MachineService<T extends Machine> {
-
     /**
      * Return the list of machines.
      * In the implementation of this method please use your MachineProducer implementations
@@ -17,8 +16,7 @@ public interface MachineService<T extends Machine> {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Machine> getAll(Class<? extends T> type);
-
+    List<T> getAll(Class<? extends T> type);
     /**
      * Fill the machines list with passed value
      * Replace the Object with parametrized value.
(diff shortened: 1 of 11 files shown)


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
```
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
...
+    void fill(List<? super T> machines, T value);
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
```suggestion
    void startWorking(List<? extends T> machines);
```
no need to use machine when we declared a class type parameter that extends it

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
-public class MachineServiceImpl {
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import core.mate.academy.model.Workable;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
```
okuzan:
```suggestion
public class MachineServiceImpl implements MachineService<Machine> {
```
fix methods accordingly
we don't need new type parameters in this class, we just implement Machine service with T as Machine

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,20 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+
```
okuzan: no empty lines after class/method definition are required

okuzan (changes requested):

okuzan (approved):
Splendid!

PR #811, opened 2024-02-13
implemented program logic using pecs rule

kriznychenko:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -15,8 +16,9 @@ public interface MachineService {
      *
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
+     *
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class type);
```
okuzan: in each method in this interface, use T class type parameter, which you defined above, instead. What was the point of having T if not using it ;)

okuzan (changes requested):

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +28,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super Machine> machines, Machine value);
```
okuzan: replace machine with T

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +28,13 @@ public interface MachineService {
...
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
okuzan: replace machine with T

okuzan (changes requested):

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +28,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, Machine value);
```
```suggestion
    void fill(List<? super T> machines, T value);
```
in methods, we should solely rely on type parameter

okuzan (changes requested):

okuzan (approved):
LGTM

PR #813, opened 2024-02-15
created MachineServiceImpl class

YuliiaNisha:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
...
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T extends Machine> {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends T> get();
```
okuzan:
```suggestion
    List<T> get();
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,55 @@
...
+        for (Machine machine : machines) {
+            if (machine instanceof Bulldozer) {
+                bulldozer.doWork();
+            }
+            if (machine instanceof Excavator) {
+                excavator.doWork();
+            }
+            if (machine instanceof Truck) {
+                truck.doWork();
+            }
```
okuzan: this should be just machine.doWork() if generics are correctly implemented

okuzan (changes requested):

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,15 +1,34 @@
...
+import core.mate.academy.service.MachineProducer;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Add some custom fields that could be only in Bulldozer
  * Do not remove no-args constructor
  */
-public class Bulldozer extends Machine {
+public class Bulldozer extends Machine implements MachineProducer<Bulldozer> {
```
okuzan: separate producers from the models
You have three types of machine, and you should have three separate producers

src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -1,6 +1,7 @@
 package core.mate.academy.model;
 
 public abstract class Machine implements Workable {
+    protected static final int MODELS_NUMBER = 3;
```
okuzan: remove this field

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,61 @@
...
+public class MachineServiceImpl implements MachineService<Machine> {
+    private Bulldozer bulldozer = new Bulldozer();
+    private Excavator excavator = new Excavator();
+    private Truck truck = new Truck();
+
+    public MachineServiceImpl() {
+    }
+
+    @Override
+    public List<? extends Machine> getAll(Class type) {
```
okuzan:
```suggestion
    public List<Machine> getAll(Class<? extends Machine> type) {
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,61 @@
...
+        List<? extends Machine> resultList = new ArrayList<>();
+        if (type == Bulldozer.class) {
+            resultList = bulldozer.get();
+        }
+        if (type == Excavator.class) {
+            resultList = excavator.get();
```
okuzan: get producer for this type, if no producer found - return empty list

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,61 @@
...
+        for (Machine machine : machines) {
+            if (machine instanceof Bulldozer) {
+                bulldozer.doWork();
+            }
+            if (machine instanceof Excavator) {
+                excavator.doWork();
+            }
+            if (machine instanceof Truck) {
+                truck.doWork();
+            }
```
okuzan:
```suggestion
            machine.doWork();
```
the point of gentrifying classes is to avoid such fragile nonuniversal code. Here we take advantage of polymorphism

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,61 @@
...
+            resultList = excavator.get();
+        }
+        if (type == Truck.class) {
+            resultList = truck.get();
+        }
+        if (type == Bulldozer.class
+                && type == Excavator.class
+                && type == Truck.class) {
+            resultList = new ArrayList<>();
+        }
```
okuzan: this condition is always false, how can type simultaneously be all three things?

okuzan (changes requested):
A lot needs to be redone. If you have troubles, don't hesitate to ask in the chat

YuliiaNisha:
I got really confused with the getAll method. From the task I understood that it should return a list of specific machine (Bulldozer, excavator or truck) or an empty list. However, when I parametrize it like this:
public List<? extends Machine> getAll(Class<? extends Machine> type)
the tests do not pass because they expect to get List<Machine>
Thanks for a hint!

olekskov (approved):

PR #821, opened 2024-02-23
(^-^)

Bohadan-Markatov:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 public interface MachineProducer {
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
```
ReamFOX: I think it would be better if you delete all these comments, because they only take up space.

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,18 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer {
```
ReamFOX: This is listed in the checklist as a bad example.
#### Do not use raw types. Always parametrize interfaces and classes if it is possible.
```java
// Bad example:

public class ExcavatorProducer implements MachineProducer {
```
```java
// Improved example:

public class ExcavatorProducer implements MachineProducer<Excavator> {
```
So use parametrized MachineProducer

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+            return new BulldozerProducer().get();
+        }
+        if (type.equals(Excavator.class)) {
+            return new ExcavatorProducer().get();
+        }
+        if (type.equals(Truck.class)) {
+            return new TruckProducer().get();
+        }
+        return new ArrayList<>();
+    }
```
ReamFOX: There's also a hint to consider creating a local variable here, but I'm not sure how critical it is.
#### Consider creating a local variable in your implementation of `getAll(Class type)` method.
```java
List<? extends Machine> machines = someProducer.get();
```

ReamFOX (changes requested):
Looks good to me😎
but tiny improvements are needed.

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,18 @@
...
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+
+    @Override
+    public List<Machine> get() {
+        List<Machine> truckList = new ArrayList<>();
+        for (int i = 0; i < 10; i++) {
```
andrii-hoienko: Magic number

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,38 @@
...
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type.equals(Bulldozer.class)) {
+            return new BulldozerProducer().get();
```
andrii-hoienko: It is possible to avoid writting return in each if statement body, you can define method level variable and populate it, also maybe switch case will be more prevefable that if/else here

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,13 @@
...
-     * Return the list of machines.
-     * In the implementation of this method please use your MachineProducer implementations
-     * See that 'Class type' is not parametrized.
-     * Consider to parametrize this Class< PARAMETRIZE_ME > with specific wildcard bounding
-     *
-     * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
-     * @return the list of machines
-     */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
andrii-hoienko:
```suggestion
    List<T> getAll(Class<? extends Machine> type);
```

src/main/java/core/mate/academy/service/TruckProducer.java
andrii-hoienko: Chech other methods as well

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,12 +1,6 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
```
andrii-hoienko, line 5: I beleive you should not remove no-args constructor and you should Add some custom fields

andrii-hoienko (changes requested):
Nice, but some improvements required

andrii-hoienko (approved):
Very nice!

PR #828, opened 2024-02-28
feat: implement MachineService

oksana-miazina:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private List<Bulldozer> list;
+
+    public BulldozerProducer() {
+        list = new ArrayList<>();
+        list.add(new Bulldozer());
+        list.add(new Bulldozer());
+    }
```
nklimovych: In my opinion, when you initialize any object with an empty constructor, you expect an empty object instead of a list of objects. This can cause misunderstandings when using your implementation. You have get() method to create a list of objects.

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class type);
```
nklimovych: I'm not sure but maybe you should use <T> instead of <Machine>:

```suggestion
    List<T> getAll(Class<? extends Machine> type);
```

nklimovych (changes requested):
It looks great, I have a few suggestions for you

src/main/java/core/mate/academy/service/BulldozerProducer.java
oksana-miazina: In general, I agree with you. But on other hand it's also incorrect to make method responsible for both operations - construction and returning.

nklimovych (approved):
Good Gob 🔥

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,23 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    private static final int NUMBER = 2;
```
yevheniisynytsia, line 9:
```suggestion
    private static final int NUMBER_OF_MACHINES = 2;
```

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,23 @@
...
+
+    private List<Bulldozer> list;
+
+    public BulldozerProducer() {
+        list = new ArrayList<>(
+                Collections.nCopies(NUMBER, new Bulldozer())
```
yevheniisynytsia, line 15: 🔥

yevheniisynytsia (approved):
Great Job! 🔥

PR #834, opened 2024-03-02
parameterized interfaces and defined methods

IgorGunder:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,13 +1,25 @@
...
+import core.mate.academy.service.MachineProducer;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Add some custom fields that could be only in Bulldozer
  * Do not remove no-args constructor
  */
-public class Bulldozer extends Machine {
+public class Bulldozer extends Machine implements MachineProducer<Machine> {
```
MentorMA:
```suggestion
public class Bulldozer extends Machine {
```

MentorMA: Create Bulldozer producer and implement there logic not in Bulldozer class

MentorMA (changes requested):

MentorMA (approved):

PR #840, opened 2024-03-06
implement Machine Producer

hlibmurphy:


src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+
+public class TruckProducer implements MachineProducer<Truck> {
+    private static final int NUM_OF_EXCAVATOR = 3;
+    private List<Truck> list = new ArrayList<>();
+
+    public TruckProducer() {
+        Random random = new Random();
```
okuzan, line 13: random should be a private field instead of a local variable

okuzan (approved):

PR #846, opened 2024-04-05
build success

MolchanovAlexander:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,9 +5,23 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
+    private int power;
```
karma-o, line 8: you have added an identical field to all 3 Machine-extending classes, this is wrong if you think about the inheritance OOP principle which helps us with such code duplications. The task description says that we can add some **type-specific** fields to those 3 classes, which means some 3 different fields that can **not** be common for all machines, because if they can - they should be a part of the `Machine` class

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,7 @@
 
 import java.util.List;
 
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<T> {
```
karma-o:
```suggestion
public interface MachineProducer<T extends Machine> {
```
because we do not want to allow some `String` or anything that is not a `Machine` here

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -2,37 +2,11 @@
 
 import java.util.List;
 
-/**
- * Parameterize this service and add its implementation in a separate class.
- */
-public interface MachineService {
+public interface MachineService<T> {
```
karma-o: same here

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+import core.mate.academy.service.producers.TruckProducer;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer<? extends Machine> producer = null;
```
karma-o: i think this can be just a wildcard or a `Machine` type, usually when you use PECS, you are declaring some methods or classes, but the variable types are getting it from where you are getting an instance of that type, and does not require using `extends`/`super`
```suggestion
        MachineProducer<Machine> producer = null;
```
or you can just return the result right away to avoid creating this variable at all

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+        if (type == Bulldozer.class) {
+            producer = new BulldozerProducer();
+        } else if (type == Excavator.class) {
+            producer = new ExcavatorProducer();
+        } else if (type == Truck.class) {
+            producer = new TruckProducer();
+        } else {
+            return new ArrayList<>();
+        }
+        return (List<Machine>) producer.get();
```
karma-o: You do not need to cast it, `get()` method should already return `List<Excavator>` or other model, which is already a `Machine` instance
```suggestion
        return producer.get();
```

src/main/java/core/mate/academy/service/producers/BulldozerProducer.java
```diff
@@ -0,0 +1,21 @@
...
+    private List<Bulldozer> machines = new ArrayList<>();
+
+    @Override
+    public List<Bulldozer> get() {
+        int randomPower = new Random().nextInt(100);
+        for (int i = 0; i < randomPower; i++) {
+            machines.add(new Bulldozer(randomPower));
+        }
+        return machines;
+    }
```
karma-o: this whole class just looks terrible: 
* you get a random 0-99 number (power can be 0?) 
* `100` should be a constant
* you create the same amount of bulldozers as you got this power
* unused `NUM_OF_MACHINES` constant
* you set the same power to all of those bulldozers (what is the point of random?)
* you do not reuse a `Random` instance as a class field, which we discussed around 2 months ago. 
* why do you have common `machines` list for a whole class? Do you realize that method `get()` will add more and more machines to it after each `get()` call? And the list that it returns will also be the same for each `get()` call?

So many and so common and silly mistakes in such a small block of code... Please think more about what you do, what your code actually does, and how does it look, 😞 
Same mistakes in other `Producer` classes

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        MachineProducer<? extends Machine> producer = null;
+        if (type == Bulldozer.class) {
+            producer = new BulldozerProducer();
+        } else if (type == Excavator.class) {
+            producer = new ExcavatorProducer();
+        } else if (type == Truck.class) {
+            producer = new TruckProducer();
```
karma-o: do not create a new producer every time, create a single instance per type, make them a class field and reuse them here

karma-o (changes requested):
Please think more deep the solution, and do some self-review of your code before pushing, because I find simple and very common mistakes quite a lot in your HW solutions

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -5,11 +5,25 @@
...
     }
 
     @Override
     public void doWork() {
         System.out.println("Truck started to work");
     }
+
+    public boolean isHasTrailer() {
+        return hasTrailer;
+    }
```
karma-o, line 24:
```suggestion
    public boolean hasTrailer() {
        return hasTrailer;
    }
```

src/main/java/core/mate/academy/service/producers/ExcavatorProducer.java
```diff
@@ -0,0 +1,25 @@
...
+    private final Random random = new Random();
+
+    @Override
+    public List<Machine> get() {
+        List<Machine> machines = new ArrayList<>();
+        int randomYear = random.nextInt(MIN_YEAR, MAX_YEAR);
+        for (int i = 0; i < NUM_OF_MACHINES; i++) {
+            machines.add(new Excavator(randomYear));
+        }
+        return List.copyOf(machines);
```
karma-o, line 23: why do we need to copy it?

karma-o (approved):
Approved with a couple more suggestions

src/main/java/core/mate/academy/service/producers/BulldozerProducer.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final int MIN_POWER = 100;
+    private final Random random = new Random();
+
+    @Override
+    public List<Machine> get() {
+        List<Machine> machines = new ArrayList<>();
+        int randomPower = random.nextInt(MIN_POWER, MAX_POWER);
+        for (int i = 0; i < NUM_OF_MACHINES; i++) {
+            machines.add(new Bulldozer(randomPower));
+        }
```
karma-o, line 22: BTW, power is still the same for all bulldozers you produce here, for it to change you need to call `nextInt()` inside the loop, to get a new number every time, because if it is outside of loop - int var `randomPower` is not going to change magically by itself
```suggestion
        for (int i = 0; i < NUM_OF_MACHINES; i++) {
            int randomPower = random.nextInt(MIN_POWER, MAX_POWER);
            machines.add(new Bulldozer(randomPower));
        }
```

PR #854, opened 2024-04-24
MateAcademy/Task/jv-pecs - Bartosz Wójcik solution

banderan:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,48 @@
...
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
+    @Override
+    public List<T> getAll(Class<? extends T> type) {
+        if (Bulldozer.class == type) {
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
+            List<T> machines = (List<T>) bulldozerProducer.get();
```
nowickimj: Try to avoid repeating code.

You can initialize generic list before `if` statement: 
```java
List<? extends Machine> machines = new ArrayList<>();
```
Then add elements for each case:
```java
machines = new BulldozerProducer().get();
```
And finally cast and return result:
```java
return (List<Machine>) machines;
```

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,14 @@
...
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozers = new ArrayList<>();
+        bulldozers.add(new Bulldozer());
+        return bulldozers;
```
nowickimj: You can use `List.of()` - which will result in single-line implementation. Output list will be immutable.
```java
return List.of(new Bulldozer());
```

nowickimj (changes requested):

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,46 @@
...
+            return (List<T>) machines;
+        } else if (type == Excavator.class) {
+            machines = new ExcavatorProducer().get();
+            return (List<T>) machines;
+        } else if (type == Truck.class) {
+            machines = new TruckProducer().get();
+            return (List<T>) machines;
+        } else {
+            return Collections.emptyList();
+        }
```
nowickimj: That's a bit different from what I've proposed.
1. Inside if/elseif body, only overwrite the `machines`.
2. Cast and return `machines` after if/elseif statement - you've already initialized an empty array list, so `Collections.emptyList()` call is not necessary.

nowickimj (changes requested):
One small fix required.

nowickimj (approved):

PR #861, opened 2024-05-07
[Java Core/PECS/jv-pecs] solution v1

ShastkivRuslan:
this is the first task after a long break, don't be afraid :)


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
```
koval-kn, line 5:
```suggestion
public interface MachineProducer<T extends Machine> {
```

koval-kn (approved):

PR #864, opened 2024-05-09
First solution

mateuszwojtkowiak:


src/main/java/core/mate/academy/model/Bulldozer.java
MateuszRostek, line 1: Getters and Setters missing in all machines.

Zrzut ekranu 2024-05-09 162311

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,21 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
```
MateuszRostek: Since `BulldozerProducer` and all created by you producers are actual implementations, it's good to add `Impl` at the end. For example `BulldozerProducerImpl`.

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,21 @@
...
+    private Bulldozer bulldozer1 = new Bulldozer("big");
+    private Bulldozer bulldozer2 = new Bulldozer("small");
+
+    private Bulldozer bulldozer3 = new Bulldozer("long");
+    private List list = new ArrayList();
+
```
MateuszRostek: You can do it inside get() method and add machines immediately when creating them.

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<T> {
```
MateuszRostek: `T` is insufficient here, it should extend something.

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,12 @@
...
     /**
      * In your implementations - create 2-3 models, add them to the list and return
      * @return - the list of models
      */
-    List<Object> get();
+    List<? extends Machine> get();
```
MateuszRostek: And then `T` is enough here:)

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T> {
```
MateuszRostek: Same here - `T` is insufficient, it should extend something.

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends Machine> type);
```
MateuszRostek: Do not hardcode `Machine` in `List<>` - use `T`. Second part is good:)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+            BulldozerProducer bulldozerProducer = new BulldozerProducer();
+            return bulldozerProducer.get();
+        } else if (type == Excavator.class) {
+            ExcavatorProducer excavatorProducer = new ExcavatorProducer();
+            return excavatorProducer.get();
+        } else if (type == Truck.class) {
+            TruckProducer truckProducer = new TruckProducer();
+            return truckProducer.get();
+        }
+        return new ArrayList<>();
```
MateuszRostek: You can also use `switch` here instead.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,43 @@
...
+
+    @Override
+    public void startWorking(List<? extends Machine> list) {
+        for (Machine machine : list) {
+            machine.doWork();
+        }
+    }
+
+    @Override
+    public void fill(List list, Object value) {
```
MateuszRostek: Remember what PECS stands for, and use it in List parameter, also think of a better name then just `list`

MateuszRostek (changes requested):
Please read the comments, also add 2 constructors in Machine class (empty and with all fields) and then in subclasses use this in constructors, passing relevant parameters and calling `super()` inside them.

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,7 +5,23 @@
...
+    private String blade;
+
     public Bulldozer() {
+        super();
+    }
+
+    public Bulldozer(String blade) {
+        super();
+        this.blade = blade;
+    }
```
MateuszRostek: Almost correct, but you don't need to call anything inside the empty construtor and in the second constructor you should also pass parameters that Machine has, then call `super()` with those params, so it would look like this:
```
public Bulldozer(String name, String color, String blade) {
        super(name, color);
        this.blade = blade;
    }
```
Same for Excavator and Truck.

src/main/java/core/mate/academy/service/BulldozerProducerImpl.java
```diff
@@ -0,0 +1,17 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducerImpl implements MachineProducer<Bulldozer> {
+    private List list = new ArrayList();
```
MateuszRostek: You can also put this inside the `get()` method.

src/main/java/core/mate/academy/service/BulldozerProducerImpl.java
```diff
@@ -0,0 +1,17 @@
...
+
+    @Override
+    public List<Bulldozer> get() {
+        list.add(new Bulldozer("big"));
+        list.add(new Bulldozer("small"));
+        list.add(new Bulldozer("long"));
```
MateuszRostek: And now you should use the updated constructor with all params here.

src/main/java/core/mate/academy/service/MachineService.java
MateuszRostek, line 1: You should also make changes in the `fill` method, do not leave `Object`.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,47 @@
...
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    private static final String bulldozerClassName = "core.mate.academy.model.Bulldozer";
+    private static final String ExcavatorClassName = "core.mate.academy.model.Excavator";
+    private static final String TruckClassName = "core.mate.academy.model.Truck";
+
+    @Override
+    public List<? extends Machine> getAll(Class type) {
```
MateuszRostek:
```suggestion
    public List<? extends Machine> getAll(Class<? extends Machine> type) {
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,47 @@
...
+
+    @Override
+    public void startWorking(List<? extends Machine> list) {
+        for (Machine machine : list) {
+            machine.doWork();
+        }
+    }
+
+    @Override
+    public void fill(List<Object> listOfMachines, Object value) {
```
MateuszRostek: It can't be just an `Object` in these, List should be parametrized with a usage of `super` and the value should be of type `Machine`.

MateuszRostek (changes requested):
Good job! Much better now, but there are still some things to improve:)

src/main/java/core/mate/academy/service/BulldozerProducerImpl.java
```diff
@@ -0,0 +1,16 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducerImpl implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List list = new ArrayList();
```
MateuszRostek, line 10: You didn't parameterize the list - also make the name more informative.

MateuszRostek: Same in all producer implementation

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,53 @@
...
+
+    @Override
+    public List getAll(Class<? extends Machine> type) {
+        /*
+        If I parametrize "Class<? extends Machine>", I can't parametrize
+        "List<? extends Machine>". I am getting this information:
+        "getAll(Class<? extends Machine>)' in 'core.mate.academy.service.MachineServiceImpl'
+        clashes with 'getAll(Class<? extends Machine>)'
+        in 'core.mate.academy.service.MachineService'; attempting to use incompatible return type"
+         */
```
MateuszRostek, line 23: I know, you should go with `List<Machine>` and then leave Class as it is now:)

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,53 @@
...
+        String typeName = type.getName();
```
MateuszRostek, line 24: You can use `type.getSimpleName()` it will give you a simpler name -> easier constants -> clearer code:)

MateuszRostek (approved):
Good job, approved, BUT please read my comments and implement them

PR #865, opened 2024-05-11
initial solution

MarcusPy:


src/main/java/core/mate/academy/service/BulldozerProducerImpl.java
```diff
@@ -0,0 +1,16 @@
...
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducerImpl implements MachineProducer<Bulldozer> {
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> machines = new ArrayList<>();
+        machines.add(new Bulldozer());
+        machines.add(new Bulldozer());
```
MateuszRostek: You have to create a constructor in Bulldozer (and in all models) with params from `Machine` (and then use `super(....)`) and unique params for each machine.

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,10 @@
 
 import java.util.List;
 
-public interface MachineProducer {
+public interface MachineProducer<MachineT> {
```
MateuszRostek: I see you used `MachineT` in a lot of places, what is it? We have `Machine` abstract class that we should use. Also it shouldn't be just `Machine` - you have to somehow use it with `extends` or `super` depending on the place you use it.

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +26,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    <T> void fill(List<T> machines, T value);
```
MateuszRostek: `List` - should be parameterized diffrently, 
Since we use parameterization at the class level, thiis `<T>` here is redundant.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,48 @@
...
+            machines = new BulldozerProducerImpl().get();
+        } else if (type == Excavator.class) {
+            machines = new ExcavatorProducerImpl().get();
+        } else if (type == Truck.class) {
+            machines = new TruckProducerImpl().get();
+        } else {
+            return List.of();
+        }
+
+        return (List<Machine>) machines;
```
MateuszRostek: You can use `switch` instead of this complicated if else, also to make code more readable you should change `List.of()` to `Collections.emptyList()`

MateuszRostek: Casting in this return statement is not necessary, you can do it differently.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,48 @@
...
+    @Override
+    public void startWorking(List<? extends Machine> machines) {
+        for (Machine machine : machines) {
+            if (machine instanceof Bulldozer bulldozer) {
+                bulldozer.doWork();
+            } else if (machine instanceof Excavator excavator) {
+                excavator.doWork();
+            } else if (machine instanceof Truck truck) {
+                truck.doWork();
+            }
```
MateuszRostek: You can just call `machine.doWork()` if you make all necessary changes in your code

MateuszRostek (changes requested):
You didn't push models (`Machine`, `Bulldozer`, etc.) or you didn't make any changes in them, please fix it! Also see some comments i provided

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,9 +5,19 @@
...
+    private int horsePower;
+
     public Bulldozer() {
     }
 
+    public Bulldozer(String brand, String model, int horsePower) {
+        this.brand = brand;
+        this.model = model;
+        this.horsePower = horsePower;
+    }
```
MateuszRostek: When you add implementation and push Machine, you should also take params that reflect machine's fields and call it's constructor using `super(...)`

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,9 +5,19 @@
  * Do not remove no-args constructor
  */
 public class Bulldozer extends Machine {
```
MateuszRostek, line 7: Provide getters/setters for all fields

MateuszRostek: Same for all subclasses of Machine

src/main/java/core/mate/academy/model/Bulldozer.java
MateuszRostek: Same for all subclasses of Machine

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<T> getAll(Class<? extends T> type);
```
MateuszRostek:
```suggestion
    List<T> getAll(Class<? extends Machine> type);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
      * @param machines - list of machines to be filled with value
      * @param value    - any object of machine sub class
      */
-    void fill(List<Object> machines, Object value);
+    void fill(List<? super T> machines, T value);
```
MateuszRostek:
```suggestion
    void fill(List<? super Machine> machines, T value);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
...
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends T> machines);
```
MateuszRostek:
```suggestion
    void startWorking(List<? extends Machine> machines);
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+        List<Machine> machines = new ArrayList<>();
+        if (type.equals(Bulldozer.class)) {
+            machines.addAll(new BulldozerProducerImpl().get());
+        } else if (type.equals(Excavator.class)) {
+            machines.addAll(new ExcavatorProducerImpl().get());
+        } else if (type.equals(Truck.class)) {
+            machines.addAll(new TruckProducerImpl().get());
+        } else {
+            return Collections.emptyList();
+        }
```
MateuszRostek: Use `switch` here!

MateuszRostek (changes requested):
Nice try, see some comments and you didn't push `Machine`/ you didn't make changes in it, so add 2 contructors in Machine.

MarcusPy:
I think all should be good now, or so I hope :)

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,9 +5,44 @@
...
+
     public Bulldozer() {
     }
 
+    public Bulldozer(String name, String color, String brand, String model, int horsePower) {
+        super();
+        this.brand = brand;
+        this.model = model;
+        this.horsePower = horsePower;
+    }
```
MateuszRostek: Right now you take 5 parameters, but you use only 3, because using empty super, `super()` just calls Machine's empty constructor, you have to pass params inside to call a proper constructor in the Machine class.
```suggestion
    public Bulldozer(String name, String color, String brand, String model, int horsePower) {
        super(name, color);
        this.brand = brand;
        this.model = model;
        this.horsePower = horsePower;
    }
```

MateuszRostek: Same for Truck and Excavator

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        List<Machine> machines = new ArrayList<>();
+        switch (type.getSimpleName()) {
+            case "Bulldozer":
+                machines.addAll(new BulldozerProducerImpl().get());
+                break;
+            case "Excavator":
+                machines.addAll(new ExcavatorProducerImpl().get());
+                break;
+            case "Truck":
```
MateuszRostek: You should make those Strings constants:)

MateuszRostek (changes requested):
You did a good job, see some minor advice and your code is ready to go:)
You still didn't make any changes in the `Machine` class, you have to add 2 constructors to it:

```java
    public Machine() {

    }

    public Machine(String name, String color) {
        this.name = name;
        this.color = color;
    }
```

MateuszRostek (approved):
Good job! Approved:)

PR #869, opened 2024-05-28
Task completed

mhvaleria:


src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -5,9 +5,17 @@
...
+    private String trackType;
+
     public Bulldozer() {
     }
 
+    public Bulldozer(String name, String color, String trackType) {
+        this.trackType = trackType;
+        setColor(color);
+        setName(name);
+    }
```
sarakhmen: why don't you call the parent constructor here? Please fix in all places

sarakhmen (changes requested):
let's improve your solution :)

koval-kn (approved):

PR #876, opened 2024-07-01
Implement MachineService interface and MachineProducer implementations

smagles:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,13 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
```
MateuszRostek, line 6: Name should suggest that `BulldozerProducer` is an implementation of something, so better a name would be: `BulldozerProducerImpl`, same for other implementations.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+import java.util.List;
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List getAll(Class<? extends Machine> type) {
```
MateuszRostek, line 16:
```suggestion
    public List<Machine> getAll(Class<? extends Machine> type) {
```

MateuszRostek (approved):
Good job! See some advice, but there are only minor mistakes, so approved!

PR #900, opened 2024-08-30
HW-pecs v.1.0

JasonTaranenko:


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +25,14 @@ public interface MachineService {
...
+    void fill(List<? super T> machines, T value);
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+
+    void startWorking(List<? extends Machine> machines);
```
```suggestion
    void startWorking(List<? extends T> machines);
```

okuzan: that's what we have T for :)

okuzan (changes requested):

Elena-Bruyako (approved):

PR #901, opened 2024-09-01
Created nessesery classes, changed Objects to PECS.

Wersebu:
Created nessesery classes, changed Objects to PECS. All tests and builds passed

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineServiceImpl.java
+++ b/src/main/java/core/mate/academy/service/MachineServiceImpl.java
@@ -10,40 +10,35 @@ import core.mate.academy.model.TruckProducer;
 import java.util.ArrayList;
 import java.util.List;
 
-/**
- * Your implementation of MachineService.
- */
-public class MachineServiceImpl implements MachineService<Machine> {
+public class MachineServiceImpl<T extends Machine> implements MachineService<T> {
 
     @Override
-    public List<Machine> getAll(Class<? extends Machine> type) {
-        MachineProducer<? extends Machine> producer;
+    public List<Machine> getAll(Class<? extends T> type) {
+        MachineProducer<? extends T> producer;
 
         if (type == Bulldozer.class) {
-            producer = new BulldozerProducer();
+            producer = (MachineProducer<? extends T>) new BulldozerProducer();
         } else if (type == Excavator.class) {
-            producer = new ExcavatorProducer();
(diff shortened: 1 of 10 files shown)


MateuszRostek (changes requested):
You should add constructors to `Machine` class:

```java
    public Machine() {

    }

    public Machine(String name, String color) {
        this.name = name;
        this.color = color;
    }
```

Then in classes that `extends` Machine you should leave one empty constructor and add another one that will look like this (of course adjusted to the relevant fields). For example in the `Bulldozer`:

```java
public Bulldozer() {
    }
    
    public Bulldozer(String name, String color, Integer fuelConsuptionPerHour) {
        super(name, color);
        this.bladeWidth = bladeWidth;
        this.cabCapacity = cabCapacity;
    }
```
Then in all producers you should create objects using the second constructor (this one with all fields), For example:

```java
    @Override
    public List<Excavator> get() {
        List<Excavator> excavator = new ArrayList<>();

        excavatorList.add(new Excavator("Exc 3000", "Blue", 60.0, 5.5));
        excavatorList.add(new Excavator("Exc 5000", "Yellow", 100.0, 8.5));
        excavatorList.add(new Excavator("Exc 800", "Yellow", 20.5, 2.5));

        return excavator;
    }
```

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,45 @@
...
+        } else if (type == Excavator.class) {
+            producer = (MachineProducer<? extends T>) new ExcavatorProducer();
+        } else if (type == Truck.class) {
+            producer = (MachineProducer<? extends T>) new TruckProducer();
+        } else {
+            return new ArrayList<>();
+        }
+
+        return (List<Machine>) producer.get();
+    }
```
MateuszRostek, line 30: Do not use over-complicated `if-else` constuction, bette to use `switch` here, for example:

```java
    @Override
        List<? extends Machine> machines;
        switch (type.getSimpleName()) {
            case "Bulldozer" -> machines = new BulldozerProducerImpl().get();
            case "Excavator" -> machines = new ExcavatorProducerImpl().get();
            case "Truck" -> machines = new TruckProducerImpl().get();
            default -> {
            }
        }
    }
```

MateuszRostek (approved):
Good job! See one minor advice and implement it!

PR #908, opened 2024-09-16
Implement Machine Service to get list of machines, replace value in list and launch machines from list.

deez-nerdy:
Add specific fields to Bulldozer, Excavator and Truck class. Create new constructors within every machine class which use this fields.

Implement MachineProducer in all type of machine: 
BulldozerProducer - to get list of bulldozers;
ExcavatorProducer - to get list of excavators;
TruckProducer - to get list of trucks;

Parametrize MachineProducer interface and methods within by using PECS:
MachineService interface - <T extends Machine> (T parameter to match name pattern);
getAll method - <? extends Machine> to get a value from collection;
fill method - <? super T> to add value to collection;
startWorking method - <? extends Machine> to get a value from collection;

Implement MachineProducer interface in MachineServiceImpl class and parametrized it:
MachineServiceImpl<T extends Machine> implements MachineService<T> (T parameter to match name pattern);

getAll method - to check type of machine and return list of this type;
fill method - to update all values in list with new value;
startWorking - to launch all machines from list;

### DIFF
--- a/src/main/java/core/mate/academy/service/MachineService.java
+++ b/src/main/java/core/mate/academy/service/MachineService.java
@@ -1,11 +1,12 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
 /**
  * Parameterize this service and add its implementation in a separate class.
  */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
 
     /**
      * Return the list of machines.
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
 
     /**
      * Fill the machines list with passed value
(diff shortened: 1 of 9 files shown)


src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -16,7 +17,7 @@ public interface MachineService {
      * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
      * @return the list of machines
      */
-    List<Object> getAll(Class type);
+    List<Machine> getAll(Class<? extends T> type);
```
Elena-Bruyako:
```suggestion
    List<T> getAll(Class<? extends T> type);
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
...
+    void fill(List<? super T> machines, T value);
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
Elena-Bruyako:
```suggestion
    void startWorking(List<? extends T> machines);
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #915, opened 2024-09-26
added solution

Oleksii-Melashchenko:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
+        } else if (type == Excavator.class) {
+            return new ArrayList<>(new ExcavatorProducer().get());
+        } else if (type == Truck.class) {
+            return new ArrayList<>(new TruckProducer().get());
```
liliia-ponomarenko: Keep if else constructions simple
Do we need else after return in if clause? If not let's not make our code longer unnecessary.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,40 @@
...
+
 /**
  * Your implementation of MachineService.
  */
-public class MachineServiceImpl {
+public class MachineServiceImpl implements MachineService<Machine> {
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(new BulldozerProducer().get());
```
liliia-ponomarenko: Better make Producers as class level variables

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -26,13 +27,13 @@ public interface MachineService {
...
+    void fill(List<? super T> machines, T value);
 
     /**
      * Call the method doWork() from each machine.
      * This method should be able to accept a list of bulldozers as well as list of trucks.
      *
      * @param machines - the list of machines
      */
-    void startWorking(List<Object> machines);
+    void startWorking(List<? extends Machine> machines);
```
liliia-ponomarenko:
```suggestion
    void startWorking(List<? extends T> machines);
```

liliia-ponomarenko (changes requested):
Good job! Let's fix a few comments ;)

Elena-Bruyako (approved):

PR #958, opened 2025-01-21
Initial commit.

Wojtek-A-JAVA:


src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozer = new ArrayList<>();
+        bulldozer.add(new Bulldozer("caterpillar", "yellow", 10, 10, "crawler"));
+        bulldozer.add(new Bulldozer("johnDeere", "green", 5, 5, "mini "));
+        bulldozer.add(new Bulldozer("case", "red", 15, 15, "hybrid "));
```
slade13: There is an extra space at the end of the string 'hybrid '. It should be 'hybrid' without the trailing space.
```suggestion
        bulldozer.add(new Bulldozer("case", "red", 15, 15, "hybrid"));
```

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+
+    @Override
+    public List<Bulldozer> get() {
+        List<Bulldozer> bulldozer = new ArrayList<>();
+        bulldozer.add(new Bulldozer("caterpillar", "yellow", 10, 10, "crawler"));
+        bulldozer.add(new Bulldozer("johnDeere", "green", 5, 5, "mini "));
```
slade13: There is an extra space at the end of the string 'mini '. It should be 'mini' without the trailing space.
```suggestion
        bulldozer.add(new Bulldozer("johnDeere", "green", 5, 5, "mini"));
```

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,17 @@
...
+
+import core.mate.academy.model.Excavator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+
+    @Override
+    public List<Excavator> get() {
+        List<Excavator> excavator = new ArrayList<>();
```
slade13: The variable name 'excavator' should be pluralized to 'excavators' to better reflect that it is a list of multiple Excavator objects, or it can be named like 'excavatorList'.

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,12 @@
...
-     * Return the list of machines.
-     * In the implementation of this method please use your MachineProducer implementations
-     * See that 'Class type' is not parametrized.
-     * Consider to parametrize this Class< PARAMETRIZE_ME > with specific wildcard bounding
-     *
-     * @param type - any class of Machine sub class. For example: Truck.class or Bulldozer.class
-     * @return the list of machines
-     */
-    List<Object> getAll(Class type);
+    void fill(List<? super T> machines, T value);
```
slade13, line 10: The `fill` method should use `List<T>` instead of `List<? super T>` for the `machines` parameter to ensure type safety and consistency with the parameterized type.

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -1,38 +1,12 @@
...
 
+import core.mate.academy.model.Machine;
 import java.util.List;
 
-/**
- * Parameterize this service and add its implementation in a separate class.
- */
-public interface MachineService {
+public interface MachineService<T extends Machine> {
+    List<Machine> getAll(Class<? extends T> type);
```
slade13: The method `getAll` should return a list of type `List<T>` instead of `List<Machine>`. This ensures that the method is correctly parameterized and returns a list of the specific machine type.

src/main/java/core/mate/academy/model/Excavator.java
```diff
@@ -1,13 +1,21 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Excavator
- * Do not remove no-args constructor
- */
 public class Excavator extends Machine {
+    private int excavatorLength;
```
slade13: The field names 'excavatorLength', 'excavatorHeight', and 'excavatorType' are prefixed with 'excavator', which is redundant since they are already part of the Excavator class. Consider renaming them to 'length', 'height', and 'type' for clarity.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        }
+        if (type == Truck.class) {
+            List<? extends Machine> machines = new TruckProducer().get();
+            return new ArrayList<>(machines);
+        }
+        return new ArrayList<>();
+    }
+
+    @Override
+    public void fill(List<? super Machine> machines, Machine value) {
```
slade13, line 30: The `fill` method should check if the list `machines` is not null before attempting to fill it to avoid potential `NullPointerException`.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+        for (int i = 0; i < machines.size(); i++) {
+            machines.set(i, value);
+        }
+    }
+
+    @Override
+    public void startWorking(List<? extends Machine> machines) {
```
slade13, line 40: The `startWorking` method should check if the list `machines` is not null before iterating over it to avoid potential `NullPointerException`.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
+
+    @Override
+    public List<Machine> getAll(Class<? extends Machine> type) {
```
slade13, line 13: The `getAll` method uses a series of `if` statements to determine the type of machine. This approach is not scalable and violates the Open/Closed Principle. Consider using a map or another design pattern to handle different machine types more elegantly. In case of 10000 machine types this method would have 10000 repeated if statements.

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -1,13 +1,20 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Truck
- * Do not remove no-field constructor
- */
 public class Truck extends Machine {
+    private int year;
```
slade13: The field 'year' should have a more descriptive name, such as 'manufactureYear', to clearly indicate what the year represents.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,42 @@
...
- */
-public class MachineServiceImpl {
+import core.mate.academy.model.Bulldozer;
+import core.mate.academy.model.Excavator;
+import core.mate.academy.model.Machine;
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MachineServiceImpl implements MachineService<Machine> {
```
slade13, line 10: The `MachineServiceImpl` class is parameterized with `Machine`, but it should be parameterized with a generic type `T` that extends `Machine` to handle different types of machines more flexibly.
```suggestion
public class MachineServiceImpl implements MachineService<T extends Machine> {
```

src/main/java/core/mate/academy/model/Bulldozer.java
```diff
@@ -1,13 +1,21 @@
 package core.mate.academy.model;
 
-/**
- * Add some custom fields that could be only in Bulldozer
- * Do not remove no-args constructor
- */
 public class Bulldozer extends Machine {
+    private int bladeLength;
+    private int ripperLength;
+    private String bulldozerType;
```
slade13: It would be better to use just `type`:
```suggestion
    private String type;
```

slade13 (changes requested):
Good job, please review my comments and add required changes, the most important thing to be changed are:
1. Various machine types handling inside `MachineServiceImpl` (if statements)
2. Correct types inside methods of `interface MachineService`

src/main/java/core/mate/academy/service/MachineServiceImpl.java
slade13: Please replace this if statement block with a loop.

slade13 (approved):
Good job, please replace if statement block with a loop inside `MachineServiceImpl `.

PR #967, opened 2025-02-20
created an implementation of the MachineProducer and MachineService i…

Roman496:
…nterfaces

### DIFF
--- a/src/main/java/core/mate/academy/model/Machine.java
+++ b/src/main/java/core/mate/academy/model/Machine.java
@@ -4,6 +4,9 @@ public abstract class Machine implements Workable {
     private String name;
     private String color;
 
+    public Machine() {
+    }
+
     public String getName() {
         return name;
     }
--- a/src/main/java/core/mate/academy/model/Machine.java
+++ b/src/main/java/core/mate/academy/model/Machine.java
@@ -4,9 +4,6 @@ public abstract class Machine implements Workable {
     private String name;
     private String color;
 
-    public Machine() {
-    }
-
     public String getName() {
         return name;
     }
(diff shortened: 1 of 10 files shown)


src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -4,6 +4,9 @@ public abstract class Machine implements Workable {
     private String name;
     private String color;
 
+    public Machine() {
+    }
+
```
liliia-ponomarenko: it's default constructor

src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -2,10 +2,8 @@
...
-public interface MachineProducer {
-    /**
-     * In your implementations - create 2-3 models, add them to the list and return
-     * @return - the list of models
-     */
-    List<Object> get();
+public interface MachineProducer<T> {
+    int LENGTH_LIST = 5;
+
+    List<? extends T> get();
```
liliia-ponomarenko:
```suggestion
public interface MachineProducer<T extends Machine> {
    int LENGTH_LIST = 5;

    List<T> get();
```

src/main/java/core/mate/academy/service/MachineService.java
```diff
@@ -2,37 +2,11 @@
 
 import java.util.List;
 
-/**
- * Parameterize this service and add its implementation in a separate class.
- */
-public interface MachineService {
+public interface MachineService<T> {
```
liliia-ponomarenko:
```suggestion
public interface MachineService<T extends Machine> {
```

liliia-ponomarenko (changes requested):
Good job! A few comments ;)

Elena-Bruyako (approved):

PR #987, opened 2025-04-15
solution 1

anpiet:


src/main/java/core/mate/academy/service/MachineProducer.java
```diff
@@ -1,11 +1,8 @@
 package core.mate.academy.service;
 
+import core.mate.academy.model.Machine;
```
slade13, line 3: The import statement for 'core.mate.academy.model.Machine' is correct, but ensure that the 'Machine' class is indeed an abstract class as required by the task description.

src/main/java/core/mate/academy/model/Machine.java
```diff
@@ -19,4 +19,13 @@ public String getColor() {
     public void setColor(String color) {
         this.color = color;
     }
+
+    public void startWork() {
```
slade13, line 23: The `startWork` method is defined but not implemented. If this method is intended to be abstract, it should be declared as such in the abstract class `Machine`. If it is meant to have a default implementation, it should contain some logic or be removed if unnecessary.

src/main/java/core/mate/academy/model/Truck.java
```diff
@@ -1,15 +1,18 @@
...
- * Do not remove no-field constructor
- */
 public class Truck extends Machine {
+
+    private int capacityInTons;
+
     public Truck() {
     }
 
+    public Truck(int capacityInTons) {
```
slade13, line 10: The constructor `Truck(int capacityInTons)` does not validate the input parameter. Consider adding validation to ensure that `capacityInTons` is a positive integer.

slade13 (approved):
Good job, please review my comments and change method `startWork` to abstract.

PR #1084, opened 2026-05-02
test

AlexPranschke:


src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+        } else if (type == Excavator.class) {
+            return new ArrayList<>(excavatorProducer.get());
+        } else if (type == Truck.class) {
+            return new ArrayList<>(truckProducer.get());
+        }
+        return Collections.emptyList();
+    }
+
+    @Override
+    public void fill(List<Object> machines, Object value) {
```
slade13: Current version still uses `Object `in `fill()` method, use generic type:

```suggestion
    public void fill(List<? super T> machines, T value) {
```

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,18 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Bulldozer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BulldozerProducer implements MachineProducer<Bulldozer> {
+
+    private List<Bulldozer> bulldozers = new ArrayList<>();
```
slade13: Do not store machine lists as fields inside producer classes. Currently each producer keeps a list as a field and adds new machines every time `get()` is called. This means the first call returns 3 bulldozers, the second call returns 6, the third call returns 9, etc. The method should create and return a fresh list each time.

src/main/java/core/mate/academy/service/ExcavatorProducer.java
```diff
@@ -0,0 +1,17 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Excavator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExcavatorProducer implements MachineProducer<Excavator> {
+    private List<Excavator> excavators = new ArrayList<>();
```
slade13: Do not store machine lists as fields inside producer classes. Currently each producer keeps a list as a field and adds new machines every time `get()` is called. This means the first call returns 3 bulldozers, the second call returns 6, the third call returns 9, etc. The method should create and return a fresh list each time.

src/main/java/core/mate/academy/service/TruckProducer.java
```diff
@@ -0,0 +1,17 @@
+package core.mate.academy.service;
+
+import core.mate.academy.model.Truck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TruckProducer implements MachineProducer<Truck> {
+    private List<Truck> trucks = new ArrayList<>();
```
slade13: Do not store machine lists as fields inside producer classes. Currently each producer keeps a list as a field and adds new machines every time `get()` is called. This means the first call returns 3 bulldozers, the second call returns 6, the third call returns 9, etc. The method should create and return a fresh list each time.

src/main/java/core/mate/academy/service/MachineServiceImpl.java
```diff
@@ -1,7 +1,41 @@
...
+    public List<Machine> getAll(Class<? extends Machine> type) {
+        if (type == Bulldozer.class) {
+            return new ArrayList<>(bulldozerProducer.get());
+        } else if (type == Excavator.class) {
+            return new ArrayList<>(excavatorProducer.get());
+        } else if (type == Truck.class) {
+            return new ArrayList<>(truckProducer.get());
+        }
+        return Collections.emptyList();
+    }
```
slade13, line 29: Use a local `List<? extends Machine>` variable in `getAll()` to avoid duplicated return logic. Current implementation works functionally, but it repeats `return new ArrayList<>(...)` in every branch, cleaner version:

```suggestion
@Override
public List<Machine> getAll(Class<? extends Machine> type) {
    List<? extends Machine> machines;

    if (type == Bulldozer.class) {
        machines = bulldozerProducer.get();
    } else if (type == Excavator.class) {
        machines = excavatorProducer.get();
    } else if (type == Truck.class) {
        machines = truckProducer.get();
    } else {
        return Collections.emptyList();
    }

    return new ArrayList<>(machines);
}
```

src/main/java/core/mate/academy/service/BulldozerProducer.java
```diff
@@ -0,0 +1,18 @@
...
+
+    @Override
+    public List<Bulldozer> get() {
+        bulldozers.add(new Bulldozer());
+        bulldozers.add(new Bulldozer());
+        bulldozers.add(new Bulldozer());
+        return new ArrayList<>(bulldozers); // kopia
```
slade13: Remove the non-English comment

slade13 (changes requested):
Please review my comments and add necessary changes.

slade13 (approved):
Good job.

